PR #31, opened 2017-11-01
Enforce permissions and allow to upload movie previews

frankrousseau:
**Problems**

* CG artists have access to project they don't work on.
* You can't upload a movie as a preview.

**Solution**

* Add permission checking on all routes depending on the situation.
* You can upload .mp4 files as preview. They are resize as 720p movies and thumbnais/previews are generated from it.


frankrousseau:
Tests are working locally. I will figure out this dependency issue later.

PR #34, opened 2017-11-13
Remove unused dependency: uwsgi

pilou-:
`uwsgi` isn't used and should be removed from `requirements.txt`.


frankrousseau:
Thx @pilou-

PR #37, opened 2017-11-14
Install dependencies using pip

pilou-:
Using pip allows to use wheel binary package and to avoid compilation steps.


frankrousseau:
Thank you, it's cleaner this way.

PR #38, opened 2017-11-14
Systemd units: add ffmpeg path, add comments

pilou-:
- when `ffmpeg` isn't in PATH, Zou fails with `NeedDownloadError` exception
- add comments about database connection


frankrousseau:
Great! Thank you

PR #42, opened 2017-12-04
Python packaging: use setup.cfg

pilou-:
Modifications:
- use [`setup.cfg`](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) (setuptools >= 30.3.0 required)
- rename `test` into `tests`
- use `click` instead of `flask_script`
- use `app.{root,instance}_path` instead of current working directory
- tests: use `app.config` instead of `config`

Test package using wheel format is available on [test PyPi instance](https://test.pypi.org/project/zou/), it can be tested using:
```
pip install --index-url https://testpypi.python.org/simple --extra-index-url https://pypi.python.org/simple zou
```


frankrousseau:
Thank you @pilou- ! I have one question. You renamed the `zou` binary. Which could break current installations. Is it really necesseray?

pilou-:
`zou` binary is still available, it is copied during installation, thanks to the [console_scripts entry_points](https://github.com/cgwire/zou/pull/42/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R78).

frankrousseau:
There is another issue. The thing is that flask script come with commands, notably runserver which runs a server on given port. It looks like there is not that kind of things in click. Should we use gunicorn now for development?

frankrousseau:
I don't merge it immediately because I have to figure out what I must change in the documentation and how to migrate existing installation.

pilou-:
> There is another issue. The thing is that flask script come with commands, notably runserver which runs a server on given port. It looks like there is not that kind of things in click. Should we use gunicorn now for development ?

`gunicorn` isn't required for development: from the source directory, the following commands can be used:
```
$ export FLASK_APP=wsgi.py
$ flask run
 * Serving Flask app "wsgi"
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
127.0.0.1 - - [05/Dec/2017 00:04:41] "GET / HTTP/1.1" 200 -
```

pilou-:
> I don't merge it immediately because I have to figure out what I must change in the documentation and how to migrate existing installation.

With the current code in master, **by default**, `THUMBNAIL_FOLDER` and `FILE_TREE_FOLDER` are relative to the current working directory.

With this pull-request, **by default**:
- `FILE_TREE_FOLDER` is relative to the application’s folder (`app.root_path`), because data are included in Python package
- `THUMBNAIL_FOLDER` is relative to the _instance folder_ (app.instance_path), because thumbnails are not included in Python package

When installed inside a **virtual environment**  `/tmp/test`:
- app.root_path is `/tmp/test/lib/python3.6/site-packages/zou/app`
- app.instance_path is `/tmp/zoutest2/var/zou.app-instance`

With **dev mode**, paths are:
- app.root_path is `/home/pilou/src/zou/zou/app`
- app.instance_path is `/home/pilou/src/zou/instance`

Note that `FILE_TREE_FOLDER` and `THUMBNAIL_FOLDER`  can be set using environment variables or a configuration file (path of this optional configuration file is set using environment variable `ZOU_SETTINGS`).

frankrousseau:
For your information, I will remove `EVENT_MAP` and put back the previous behaviour. The idea is to allow studios to write their own event handlers. That's why event map is built from the content of the EVENT_HANDLERS folder.

pilou-:
> The idea is to allow studios to write their own event handlers.
This is still possible: `EVENT_MAP` need to be defined in the custom Python configuration file. Environment variable `ZOU_SETTINGS` must be set with the path of the custom Python configuration file.

frankrousseau:
These changes were included in #45 . Thank you a lot @pilou- !

PR #59, opened 2018-02-01
task comment can use provided person_id as comment author

flablog:
**Problem**
- Posting a comment through Gazu providing a specific user_id as author (for automated enviromnent without a proper user login for example) was not working

**Solution**
- allow the post comment to use an optional person_id in the data provided, and use it as person instead of currently loggued account


frankrousseau:
Great thank you! Next time to protect your code from future changes, I recommend you to add a test.

flablog:
Oh yes, I need to learn that part more..
Including for gazu

PR #86, opened 2018-06-20
Only warn about non-valid email.

tokejepsen:
**Problem**
When creating an admin user, the email has to be valid.

**Solution**
Instead failing to create an admin user with an invalid email. Just warn about the validity of the email.

### DIFF
--- a/zou/cli.py
+++ b/zou/cli.py
@@ -67,6 +67,10 @@ def create_admin(email):
 
     try:
         auth.validate_email(email)
+    except auth.EmailNotValidException:
+        print("WARNING: Email is not valid.")
+
+    try:
         password = auth.encrypt_password("default")
         persons_service.create_person(
             email,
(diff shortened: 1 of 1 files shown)


frankrousseau:
Thank you for the contribution. But to be honest, I'm not confortable with the idea of allowing non valid email as login.

tokejepsen:
I can see why you might not want that. Maybe there is a different solution to this? Like maybe forcing an email through?

The issue originated when I was trying to build the [docker image](https://github.com/cgwire/cgwire) locally, and the default ```admin@example.com``` specified [here](https://github.com/cgwire/cgwire/blob/master/init_zou.sh#L12), could not be validated.

frankrousseau:
Because at some point, I want to add emailing feature to CGWire. Without a valid email, it won't be able to work properly. 

From what you mention, I think the real issue is that admin@example.com should not be considered as invalid.

tokejepsen:
> From what you mention, I think the real issue is that admin@example.com should not be considered as invalid.

Ahh, so maybe we should have an exception for ```admin@example.com``` instead of allowing all emails?

PR #124, opened 2019-01-28
Docs

flablog:
**Problem**
- Some typos and errors were stil in the documentation : 
    - Home folder not created
    - wrong config file name used (/etc/sysctl.conf)
    - missing password argument while creating the admin user
- The example needs interactive prompts from the user to setup the database password

**Solution**
- Fixing the typos
- Adding an alternative way of setting the password using a simple command

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -135,6 +135,11 @@ Enter it again:
 
 Then exit from the postgres client console.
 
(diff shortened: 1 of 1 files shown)


frankrousseau:
Wow awesome! Thank you @flablog

PR #135, opened 2019-02-22
Fixing formality / grammar / punctuation of documentation

cs-student101:
All the changes were applied to the README.md file. Made documentation more formal by replacing instances of "we", "your", etc. The author section is left the way it is to allow for better expressiveness from the authors. Fixed a few grammar and punctuation mistakes.

**Problem**
Various mistakes that may make it more difficult for readers to understand the documentation.

**Solution**
Fixed some of the grammar and punctuation mistakes for better clarity, along with other changes to README.md

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,12 +4,10 @@
 Zou is the memory of your CG production
 ---------------------------------------
 
-Zou is an API that allows to store and manage the data of your CG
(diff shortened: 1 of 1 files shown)


frankrousseau:
Thank you @cs-student101!

PR #145, opened 2019-04-02
fix check_working_on_entity

pcharmoille:
**Problem**
Permission checking had issues when checking if user were assigned to a task on a given entity

**Solution**
Fix permission checking method `check_working_on_entity`


frankrousseau:
Thank you @pcharmoille!

PR #159, opened 2019-06-11
Add DOMAIN_PROTOCOL env var

g-LuL:
**Problem**
If domain is only http (I know it's bad ;) ), link in password recovery email was https. So the link was wrong.

**Solution**
Use a DOMAIN_PROTOCOL environment variable to handle https or http in the link.


frankrousseau:
Awesome! Thank you!

PR #172, opened 2019-08-14
fix ProductionTeamResource get

pcharmoille:
**Problem**
Could not get a project team because `get_project` method missed an argument

**Solution**
Add missing argument and use `get_project_raw`


frankrousseau:
Great, thank you!

PR #173, opened 2019-08-15
Fix output files from entity request

aboellinger:
Fix bug returning all asset instances output files for the provided entity.

**Problem**
Using this method to get output types from an entity owuld also return all the the output types from instances of that entity.

**Solution**
Add a filter to the request to ignore records where `asset_instance_id` is defined


frankrousseau:
Thx for the fix!

PR #176, opened 2019-08-23
Episode task

aboellinger:
**Problem**
* The backend provides no routes to get tasks related to an episode (such as _script_, _storyboard_, _animatic_, _edit_, etc.)
* The routes for sequence and scene tasks are not tested

**Solution**
* Add missing routes 
* Add missing tests


zou/app/services/shots_service.py
```diff
@@ -266,6 +266,81 @@ def get_shots_and_tasks(criterions={}):
     return list(shot_map.values())
 
 
+def get_entities_and_tasks(criterions={}):
```
frankrousseau: Can you move this in the entities_service?

zou/app/services/tasks_service.py
```diff
@@ -192,29 +192,38 @@ def get_tasks_for_asset(asset_id):
...
+    episode = shots_service.get_episode_raw(episode_id)
+    return get_task_dicts_for_entity(episode.id)
+
+
 def get_task_dicts_for_entity(entity_id):
     """
     Return all tasks related to given entity. Add extra information like
     project name, task type name, etc.
     """
-    query = Task.query.order_by(Task.name) \
```
frankrousseau: If you can keep the previous code style, it's better. It would be more coherent with the rest.

zou/app/services/shots_service.py
aboellinger: Done

zou/app/services/tasks_service.py
aboellinger: Done

frankrousseau:
Thx!

PR #182, opened 2019-09-24
Entity and instance output files

pcharmoille:
**Problem**
- There was no route to get all output files for an instance or an entity
- Routes to get last output files for an instance or entity were to slow and return type was not easy to handle

**Solution**
- add routes to get all output files for an instance or an entity
- routes to get last output files for an instance or entity return a list

Note : I need to submit a linked PR in Gazu


frankrousseau:
Hello Paul,

Thank you for your PR. Can you give you more explanation of what was wrong with the last output files functions? Especially about the format returned.

pcharmoille:
Files were grouped by output type and names. 
Since we could wanted to group files by any other criteria it was complicated to ungroup and group again on client side.
Furthemore we add more criterions and it seemed biased to only group by two of them.

Also we gained a lot of time using join queries (from 10s a request to less than a second)
If it's ok for you I'll correct the tests.

pcharmoille:
Linked PR in Gazu : https://github.com/cgwire/gazu/pull/73

tests/files/test_route_output_files.py
```diff
@@ -94,26 +94,12 @@ def test_get_last_output_files(self):
...
-        self.assertEqual(
-            output_files[str(self.render_id)]["variant-1"],
-            self.output_file_render_2.serialize()
-        )
+
+        assert(self.output_file_geometry.serialize() in output_files)
+        assert(self.output_file_cache.serialize() in output_files)
+        assert(self.output_file_texture.serialize() in output_files)
+        assert(self.output_file_render_1.serialize() in output_files)
+        assert(self.output_file_render_2.serialize() in output_files)
```
frankrousseau: Can you use `self.assertIn(a, b)` instead?

tests/files/test_route_output_files.py
```diff
@@ -501,9 +487,8 @@ def test_get_last_instance_outputs(self):
                 shot_id
             )
         )
-        self.assertEqual(
-            result[self.cache_type_id]["main"]["id"],
-            output_file["id"]
+        assert(
+            output_file["id"] in [f['id'] for f in result]
```
frankrousseau: Same here.

tests/services/test_files_service.py
```diff
@@ -204,14 +204,28 @@ def test_get_last_output_files_for_entity(self):
...
+
+        # test last geometry file revision
+        last_file = [
+            f
+            for f in last_output_files
+            if (
+                f['output_type_id'] == str(geometry.id)
+                and f['name'] == geometry_file.name
+            )
+        ][0]
```
frankrousseau: Can you find a more straightforward test? It's a good practice to put at least logic as possible in tests (otherwise you must test the test too).

zou/app/services/files_service.py
```diff
@@ -387,82 +387,194 @@ def get_last_output_revision(
...
+            OutputFile.output_type_id == statement.c.output_type_id,
+            OutputFile.name == statement.c.name,
+            OutputFile.representation == statement.c.representation,
+            OutputFile.revision == statement.c.MAX
+        )
+    )
 
-        if output_type_id not in result:
-            result[output_type_id] = {}
+    # Filter by specified arguments
```
frankrousseau: It may be better to apply the filter before the join (less joins to perform).

zou/app/services/files_service.py
```diff
@@ -387,82 +387,194 @@ def get_last_output_revision(
...
+        OutputFile.name,
+        OutputFile.representation,
+        func.max(OutputFile.revision).label("MAX")
+    ).group_by(
+        OutputFile.task_type_id,
+        OutputFile.output_type_id,
+        OutputFile.name,
+        OutputFile.representation
+    )
+    query = query.filter(OutputFile.entity_id == entity_id)
```
frankrousseau: Can you chain the call and get only one assignation?

tests/files/test_route_output_files.py
pcharmoille: done

tests/files/test_route_output_files.py
pcharmoille: done

tests/services/test_files_service.py
pcharmoille: something like this ?
```python
        # test last geometry file revision
        last_output_files = files_service.get_last_output_files_for_entity(
            self.asset.id,
            output_type_id=str(geometry.id),
            name=geometry_file.name
        )
        self.assertEqual(len(last_output_files), 0)
        self.assertEqual(last_output_files[0]['revision'], 5)
```

frankrousseau: Yes!

zou/app/services/files_service.py
frankrousseau: Discussed live. It's ok like this.

frankrousseau:
Thank you for this addition!

PR #185, opened 2019-10-09
entity working file route

pcharmoille:
**Problem**
There is no route to get all the working files for an entity

**Solution**
Add a route to get all the working files for an entity


pcharmoille:
If it's approved I'll implement the tests

frankrousseau:
Everything looks good! I wait for your tests.

PR #188, opened 2019-10-11
Rebase code from Unit Image

pcharmoille:
- fix short name import from Shotgun
- add env var to set sql alchemy pool size and overflow
- handle templates for tasks in episodes
- fix get_last_output_files_for_entity query filter


frankrousseau:
Great, thank you!

PR #190, opened 2019-10-11
Prevent event-handlers to crash zou (log error instead)

aboellinger:
**Problem**
A bug introduced by an event handler can cause Zou to crash

**Solution**
Use a try catch to catch and log errors from the event handler.

### DIFF
--- a/zou/app/utils/events.py
+++ b/zou/app/utils/events.py
@@ -1,8 +1,11 @@
 from collections import OrderedDict
 
+from flask import current_app
+
 from zou.app.stores import publisher_store
 from zou.app.models.event import ApiEvent
 from zou.app.utils import fields
+
 
 handlers = {}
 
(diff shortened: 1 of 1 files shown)


zou/app/utils/events.py
```diff
@@ -69,7 +69,14 @@ def emit(event, data={}, persist=True):
             from zou.app.stores.queue_store.job_queue import enqueue
             enqueue(func.handle_event, data)
         else:
-            func.handle_event(data)
+            try:
+                func.handle_event(data)
+            except Exception:
+                import traceback
```
frankrousseau: Can you handle exception this way:

```
from flask import current_app

...

current_app.logger.error(Error handling event, exc_info=1)
```

The `exc_info` parameter will include the stacktrace.

frankrousseau:
Great!, thank you!

PR #193, opened 2019-10-14
Fix regressions in Unit Image codebase

aboellinger:
**Problem**
Some expected attributes were not returned anymore, causing regressions in Unit Image deployment tests.

**Solution**
Make sure those attributes find their way back in the entities returned by zou.


frankrousseau:
For the shot list, can you add the missing fields instead of adding everything? It has impact on the Vue.js frontend performances.

aboellinger:
> For the shot list, can you add the missing fields instead of adding everything? It has an impact on the Vue.js frontend performances.

Done

frankrousseau:
Thx

PR #195, opened 2019-10-28
Fix filtering results in last output file for entity resource

aboellinger:
**Problem**
`files_service.get_last_output_files_for_entity` offers additional filtering arguments, but they are not exposed to the REST API.

**Solution**
Get optional arguments from the request arguments, and pass them to `files_service.get_last_output_files_for_entity`


frankrousseau:
Thx

PR #196, opened 2019-11-03
Fix Active Directory Sync

hasusuf:
**Problem**
- Syncing Disabled accounts, User accounts, and Computer accounts

**Solution**
- Adapting the LDAP query to only fetch User accounts and exclude any none-active accounts

### DIFF
--- a/zou/app/utils/commands.py
+++ b/zou/app/utils/commands.py
@@ -170,7 +170,7 @@ def sync_with_ldap_server():
             "mail",
             "thumbnailPhoto",
         ]
-        conn.search(LDAP_BASE_DN, "(objectclass=person)", attributes=attributes)
+        conn.search(LDAP_BASE_DN, "(objectCategory=person)", attributes=attributes)
         return [
             {
                 "first_name": clean_value(entry.givenName),
--- a/zou/app/utils/commands.py
+++ b/zou/app/utils/commands.py
@@ -169,6 +169,7 @@ def sync_with_ldap_server():
             "sAMAccountName",
             "mail",
             "thumbnailPhoto",
+            "userAccountControl"
         ]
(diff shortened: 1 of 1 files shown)


zou/app/utils/commands.py
```diff
@@ -170,7 +170,7 @@ def search_ad_users(conn, excluded_accounts):
             "mail",
             "thumbnailPhoto",
         ]
-        conn.search(LDAP_BASE_DN, "(objectclass=person)", attributes=attributes)
+        conn.search(LDAP_BASE_DN, "(objectCategory=person)", attributes=attributes)
```
frankrousseau: Don't you think we should use `objectclass=user` instead? 
https://social.technet.microsoft.com/wiki/contents/articles/5392.active-directory-ldap-syntax-filters.aspx

hasusuf: I tried to use `objectclass=user` at first, but it didn't work as expected. What works with me was `objectCategory=person`

hasusuf: Maybe we need to use both of them? I will try that and get back to you.

frankrousseau: Don't worry, it was just a question. Your proposal is already an improvement. We can merge it as it is.

hasusuf: Sorry @frankrousseau for my late response, it was a pretty backed week!

I tried to mix `objectCategory` with `objectclass=user` as the documentation said, it fails with `LDAPInvalidScopeError('invalid scope type')`  🤷🏻‍♂️

However, I noticed that as long as we are filtering with `userAccountControl` it will always return a `User object`

If you are happy we that please feel free to merge.

frankrousseau: Ok great, let's merge it! Thank you for this contribution!

PR #200, opened 2019-11-18
Added routes to get tasks and task types for user and sequence.

rvagner42:
**Problem**
There was no routes to get the sequences tasks and task types for the current user.

**Solution**
I added the routes.


frankrousseau:
Awesome, thank you @rvagner42!

PR #201, opened 2019-11-19
Deactivate disabled AD users when syncing

mathbou:
**Problem**

-  Disabled AD users are not deactivated when syncing users

**Solution**

- Improve LDAP sync by taking *active* parameter into account


frankrousseau:
Awesome! Thank you @mathbou

PR #202, opened 2019-11-21
Update development.md

GoOz:
Fix cli command to launch postgres docker

**Problem**
The written command doesn't work well and the docker image crashes.

**Solution**
This new commands works perfectly now.


frankrousseau:
Thx!

PR #203, opened 2019-11-21
Add jsonb data column to task.

g-LuL:
**Problem**
data field was missing in task

**Solution**
add data in task


frankrousseau:
The migration script is wrong but we will fix it soon.

frankrousseau:
Thank you for the PR!

g-LuL:
I was not pretty sure of my migration files.  
In the migration script, i saw lines that did not refer to my modification... But I didn't know how to fix this.
Anyway, first time doing this and i didn't found documentation about that process. Is there any somewhere ?

frankrousseau:
You did well. It's just that it seems that a migration file is missing. 

If you want to understand better what is happening, I recommend you to read the Alembic documentation (not related with the CG file format). 
https://alembic.sqlalchemy.org/en/latest/

PR #220, opened 2020-01-14
Fix data extraction for Shotgun tasks

pcharmoille:
**Problem**
Importing tasks from Shotgun sometimes lead to wrong link in task type.
As multiple task types (steps in Shotgun) can have the same name, we
need to identify them from their Shotgun id.

**Solution**
identify Shotgun steps using their shotgun id.


frankrousseau:
Great thank you!

PR #224, opened 2020-02-17
fix import

pcharmoille:
**Problem**
an old method/object was imported

**Solution**
fix the imported object


frankrousseau:
Great, thanks @pcharmoille!

PR #229, opened 2020-02-25
add comment id to preview-file:create event data

pcharmoille:
**Problem**
Currently there is no way to access a comment from a preview while dealing with preview creation.

**Solution**
Add the comment id to the data sent at preview file creation.

Another way might would be the implementation of a method to retrieve the comment from a preview (and the other way around).


frankrousseau:
It makes sense. Thank you for this addition. 

"a method to retrieve the comment from a preview" would be great too. Let's keep that in mind.

PR #230, opened 2020-02-25
remove duplicated method

pcharmoille:
**Problem**
`get_task_statuses` is defined two times

**Solution**
remove duplicate


frankrousseau:
Thx

PR #231, opened 2020-02-27
readd asset attributes in get_full_assets

pcharmoille:
**Problem**
We need `entities_in` field when retrieving all fields of an asset.
I don't see any other way to retrieve it from any route.
I am not sure though why it has been removed so @frankrousseau might have to check it out. The related commit is  https://github.com/cgwire/zou/commit/8eddf59605f7aa4cc013d233b9239c34b6904c73

**Solution**
do not delete attributes.

### DIFF
--- a/zou/app/services/assets_service.py
+++ b/zou/app/services/assets_service.py
@@ -332,10 +332,7 @@ def get_full_asset(asset_id):
         asset["asset_type_id"] = asset_type["id"]
         asset["asset_type_name"] = asset_type["name"]
         del asset["source_id"]
-        del asset["shotgun_id"]
         del asset["nb_frames"]
-        del asset["parent_id"]
-        del asset["entities_in"]
         asset.update(assets[0])
         return asset
     else:


frankrousseau:
It's something I removed because I was not sure it was used. IT's ok to put back `entities_in` but do you need these ones?

```
        del asset["source_id"]
        del asset["shotgun_id"]
        del asset["nb_frames"]
        del asset["parent_id"]
```

pcharmoille:
Yes, we use `source_id`, `shotgun_id` and `parent_id`.
The only one don't use on our side is `nb_frames`.
But I don't know why we would be deleting base attributes when we want the full asset. What would be the use case ?

frankrousseau:
The thing is that entity is used by the api, but we want to expose only fields that are related to asset route in assets routes. Do you use source_id for the episode? 
We can at least remove nb_frames.

pcharmoille:
My bad, we don't use `source_id` for the moment, but we do use `parent_id` to get the episode.
I could see `nb_frames` used for fx simulations but we don't use it for the moment.

So we could remove nb_frames, source_id only if it's ok for you

frankrousseau:
Let's do that!

pcharmoille:
done

PR #233, opened 2020-03-06
Improve video encoding settings

maxdudik:
**Problem**
Current video encoding settings are not optimal for file size, especially for such content as storyboards, playblasts and renders with low in-frame activity.

**Solution**
Change ffmpeg's rate control mode to Constant Rate Factor instead of Constant Bitrate.

I've tested several ffmpeg options on various types of content and found that CRF value of 15 gives pretty identical quality and reduces overall size almost twice.

### DIFF
--- a/zou/app/utils/movie_utils.py
+++ b/zou/app/utils/movie_utils.py
@@ -68,7 +68,7 @@ def normalize_movie(movie_path, fps="24.00", width=None, height=1080):
             pix_fmt="yuv420p",
             format="mp4",
             r=fps,
-            b="28M",
+            crf="15",
             preset="medium",
             vcodec="libx264",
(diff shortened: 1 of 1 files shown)


frankrousseau:
Hello,
Thank you for this PR. Can you still build playlist with this setting?

maxdudik:
Hi. We are not using playlist feature yet, so i can't say

maxdudik:
Now tested on my dev deployment and it works

frankrousseau:
Ok. We have to make tests on our side. It could lead to a lot of problems. So we want to be sure it works as expected.

frankrousseau:
I did some tests. The videos look similar. Let's try!

PR #234, opened 2020-03-06
emit asset-type event only at creation

pcharmoille:
**Problem**
`asset-type:new` event is emitted even though it is already existing.

**Solution**
Move `asset-type:new` at asset type creation.


frankrousseau:
Great, thank you!

PR #235, opened 2020-03-10
Fixed problem with slugify lowering the case of asset instance names

rvagner42:
**Problem**
The `build_asset_instance_name()` called the `slugify()` function and it lowered the case of the asset name, which was problematic for assets that contain upper case letters, since it created instances with a name different from the asset's name.

**Solution**
Use the `lowercase` keyword in the `slugify()` function so it does not change the case when building an asset instance name.

### DIFF
--- a/tests/services/test_breakdown_service.py
+++ b/tests/services/test_breakdown_service.py
@@ -103,7 +103,6 @@ class BreakdownServiceTestCase(ApiDBTestCase):
         self.assertEqual(len(casting[self.forest_id]), 2)
         self.assertEqual(len(casting[str(self.asset.id)]), 1)
 
-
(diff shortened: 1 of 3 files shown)


frankrousseau:
Thx. Can you add a test for this?

rvagner42:
Will do.
However, I just noticed that the fixture assets in the tests already have uppercase letters in their names (`"Tree"` and `"Rabbit"`), do you want me to make them lowercase instead and add a fixture asset `uppercase_name` or would you rather I keep them as is and fix all the `assertEquals` on the instances' names ?

rvagner42:
I ended up simply fixing the tests to take the uppercase letter of the assets' names.

PR #242, opened 2020-04-07
rework get_last_working_files_for_task

pcharmoille:
**Problem**
`get_working_files_for_task` was slowing down because of serialization
of all the files.

**Solution**
Only query last files using a subquery gathering maximum revision for
each task id and name


frankrousseau:
Thx!

PR #251, opened 2020-05-12
add the a file status parameter when retrieving output files

pcharmoille:
**Problem**
No possibility to filter output files using their file status

**Solution**
Add a file status id parameter for methods retrieving output files


frankrousseau:
Great, thank you!

PR #252, opened 2020-05-12
Fixed: generating a playlist-render(movie) fails when a movie has no audio, 

kguyaux:
**Problem**
Generating a playlist-render fails when one or more of the moviefiles have no audio.

**Solution**
I changed some things in `utils.movie_utils.build_playlist_movie` to fix that. It checks if a source has no audio. If that is the case, it adds an empty stream. (`ffmpeg.input('anullsrc', format='lavfi', t=0.01)`)

I have included a test but not the test-sources, because I feared the commit became to large with examplemovies and such.
We have it running in production for more than a month and as far as I know the above problem is now gone.
But please check it out first :-)


kguyaux:
Aha! I see that it is already fixed! Nevermind, this pull request will self-destruct in 3..2..1..

frankrousseau:
A what a pity I spent so much time to fix this. I think you can push the tests with related files. That would be great.

kguyaux:
Hi Frank, that's a bummer, if this whole coronacrisis didn't came along, I would have created a pull-request earlier. But to make it up I'll post the testfiles before end of tomorrow.

PR #255, opened 2020-05-20
Added section on authentication on the REST API.

ricekab:
**Problem**
During testing of the API, the authentication mechanism is not explained in the Zou docs.

**Solution**
I added a short section in `api.md` detailing the authentication mechanism (using curl). I believe this should be sufficient for someone to get their own setup going if they are not using Gazu.

*Sidenote*

Why are the docs in a completely separate branch? Just curious as to why this decision was made.


frankrousseau:
Awesome, thank you Kevin!

PR #256, opened 2020-05-22
import departments from shotgun

pcharmoille:
**Problem**
Link between user and departments were not imported

**Solution**
Add a link between user and department when a `department` field is filled.


frankrousseau:
Awesome! Thank you

PR #258, opened 2020-05-25
list all time spents when getting time spent on a task

pcharmoille:
**Problem**
When getting time spent on a task, only the last `time_spent` entry was returned

**Solution**
list all `time_spent` entries when getting time spent on a task


frankrousseau:
Great thank you!

frankrousseau:
There is an issue. You didn't add the related route.

frankrousseau:
And you added a test in tasks service while you were testing route. The right place for this is the `tests/tasks` folder.

PR #259, opened 2020-06-01
do not group by file status when retrieving last output files

pcharmoille:
**Problem**
When getting last output files, files were grouped by status.
If the status was not specified, this could lead to a result where 2 or more files would share the exact same parameters but only the status would have differ.

**Solution**
Do not group by file status. Only filter results where it it not matching the specified value.


frankrousseau:
Ok thank you!

PR #260, opened 2020-06-02
group and filter ONLY if file_status is specified

pcharmoille:
**Problem**
I messed up with the previous pr : https://github.com/cgwire/zou/pull/259
Let's say we have 3 files.
`v1` and `v2` are of status `s1`
`v3` is of status `s2`
Lets say I'd like the last output file of status `s1` wich is `v2`.
Previously the query would get the higher version, and check if match a status if specified. If not it was discarded leading the user to not get the right file.
Now, ONLY if the status is specified, the files are grouped by status, and then groups are filtered, leading the user to get the right file.

**Solution**
Only group otuput files if the file status is specified.

### DIFF
--- a/zou/app/services/files_service.py
+++ b/zou/app/services/files_service.py
@@ -498,32 +498,43 @@ def get_last_output_files_for_entity(
     params.
     """
     # Query maximum revision for each possible arguments.
-    # Only query, group and filter by file status if specified.
-    # Otherwise this could lead to different groups of files sharing the same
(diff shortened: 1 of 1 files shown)


frankrousseau:
Can you make the code smaller by applying the necessary `group_by` to the query only if it's needed? I'm not sure that SQLAlchemy allows that but it should be worth the try.

pcharmoille:
It does !

PR #285, opened 2020-10-19
Update log_rotation.md

LarsCawley:
**Problem**
I found the log rotation instructions did not work for me

**Solution**
If it's useful I have changed these instructions to the procedure that worked for me.  I changed the log paths to match the paths given in the Zou install guide, and used a different pid file creation method.


frankrousseau:
Great thank you!

PR #287, opened 2020-10-20
Update log_rotation.md

LarsCawley:
Changed the file permissions of the newly created log file to match the original, from 640 to 644.

**Problem**
New log file's permissions were chmod 640, but original is 644

**Solution**
Edited the example log rotate script so that the the permissions are the same


frankrousseau:
Thx @LarsCawley

PR #288, opened 2020-10-21
Update log_rotation.md

LarsCawley:
Tidying inconsistent /run directory paths

**Problem**
run directory paths were using a mix of /var/run and /run

**Solution**
removed /var from paths


frankrousseau:
Thx @LarsCawley

PR #296, opened 2020-11-19
Add done to the evolution event statistics

aboellinger:
⚠️ This PR is intended to be a discussion support and has not been properly tested

**Problem**
In the retake views, the edge case of tasks being both retakes and done cannot be handled correctly.

**Solution**
Add the take elements were validated at to the take statistics, so that we can count thoses tasks twice, both as retakes and as validated.

### DIFF
--- a/zou/app/services/stats_service.py
+++ b/zou/app/services/stats_service.py
@@ -33,6 +33,12 @@ DEFAULT_RETAKE_STATS = {
 }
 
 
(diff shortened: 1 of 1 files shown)


frankrousseau:
Great, thank you @aboellinger. It will improve the retake count for the edge case you describe. I will handle the Kitsu update and the test fixes.

PR #304, opened 2021-01-17
[auth/person] Add feature to protect(hide) some fields on serializer

fananimi:
**Problem**

Related issue: [Encrypted password shown on comment api response #614](https://github.com/cgwire/kitsu/issues/614)

**Solution**

Create `Meta` class on base models and put `exclude_serializes` on its property. The `exclude_serializes` used to protect/hide some fields like `password` field on person models. Then the protected fields are not serialized and not returned in any api/json response.

### DIFF
--- a/zou/app/blueprints/auth/resources.py
+++ b/zou/app/blueprints/auth/resources.py
@@ -135,7 +135,6 @@ class AuthenticatedResource(Resource):
     def get(self):
         try:
             person = persons_service.get_person_by_email(get_jwt_identity())
-            del person["password"]
             organisation = persons_service.get_organisation()
             return {
                 "authenticated": True,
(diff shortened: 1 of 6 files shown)


frankrousseau:
Thank you a lot @fananimi for this PR. But I'm a little bit concerned. You replace some methods that take advantage of the cache with raw functions. I prefer not to change that behavior. Zou already generates a lot of requests for Postgres. I would prefer to keep the previous behaviour.

fananimi:
Thanks @frankrousseau  for your feedback. 

Let me changed and re-writing the code.

Cheers!

frankrousseau:
I finally fixed that in another commit in a way that doesn't impact the data caching. Thank you for your help!

PR #312, opened 2021-02-05
Move zou.app.utils.movie_utils to zou.utils.movie

pilou-:
Move `zou.app.utils.movie_utils` in a separate distribution package. This move will allow to reuse `movie_utils` module without the need for installing all the `zou` dependencies.

* `movie_utils`: remove Zou dependency (`flask`, `zou`)
* use a namespace package for `zou`, otherwise `zou.utils` couldn't be a namespace package and couldn't be a separate distribution package ([see](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages))


frankrousseau:
Sorry but I'm not sure about this one. It is going to make the release process more complex. We are a small team, we cannot afford to add complexity on this aspect.

pilou-:
Rebased on top of #316.

Multimedia files can be processed with `ffmpeg/ffmpeg-python` either from the *local* `rq` worker (the current implementation which requires `rq` worker to be executed where the web app is executed) or from a remote worker (our goal). In both case the same operations are performed.

Therefore:
* either the remote worker needs the whole Zou application with all the Python dependencies. This is what do the jobs running within the `rq` worker ([1](https://github.com/cgwire/zou/blob/9b86135814a70f9d21e4a79ec5b97bd56625aece/zou/app/blueprints/playlists/resources.py#L123-L127), [2](https://github.com/cgwire/zou/blob/9b86135814a70f9d21e4a79ec5b97bd56625aece/zou/app/services/playlists_service.py#L497)).
* or the Python code using `ffmpeg/ffmpeg-python` (`zou.app.utils.movie_utils`) is moved in a separate distribution package using [namespace packages](https://packaging.python.org/guides/packaging-namespace-packages)

There are some alternatives: for example the duplication of the `zou/app/utils/movie_utils.py` code.

The first choice doesn't make much sense: the remote worker doesn't need the whole `zou` application to process the multimedia files.
The second choice makes the release process a little longer but not really more complex: `bdist_wheel` and `twine` commands are already used to release the `zou` Python package. Besides I guess `zou.utils` will not be updated as often as `zou`.

frankrousseau:
Thank you @pilou- for this pull request. But we can't afford to make our release cycle more complex. We are a very small team. I think it's ok to bring all the zou dependencies to use only a small subset of them. It doesn't look clean but I prefer a simpler release cycle.

pilou-:
I don't understand how this update will make the release cycle too complex. Would you plan a talk about that ?
Putting aside the new package, would that be ok to only move `zou.app.utils.movie_utils` to zou.utils.movie` ?

frankrousseau:
Yes, we can talk about it but public discussions are great too for open source projects. 

About the `movie_utils` package, it's ok to rename it. I named it like that because I was afraid of conflicts with a variable named `movie`.

pilou-:
> Yes, we can talk about it but public discussions are great too for open source projects.

Then you can explain here the current release cycle and how/why it would be too complex with this update.

frankrousseau:
Ok. So, when you have a single package to deploy, you make your modifications and commit. Then you create a version commit and you publish the package to pypi.

If you have two packages. You have to check every time that you haven't change the first package. If you changed it, you have to publish it then update its version in the main package. Once done, you can publish the main package.

pilou-:
I removed the packaging related change.

pilou-:
Rebased on `master` branch.

PR #313, opened 2021-02-05
Allow to disable mails sent via flask-mail

pilou-:
Allow to [suppress emails](https://pythonhosted.org/Flask-Mail/#unit-tests-and-suppressing-emails). This can be enabled while developing.


frankrousseau:
I think it's better to use MAIL_DEBUG=1 but thank you anyway!

PR #315, opened 2021-02-05
Refactor: always call end_build_job

pilou-:
**Requires #314**

Call `end_build_job` in `finally` block: don't require the caller of `build_playlist_movie_file` to call `end_build_job`.


frankrousseau:
Thank you!

PR #320, opened 2021-02-15
Improve normalization, use remote worker (nomad)

pilou-:
I am testing the whole patch, hence the `WIP` label.
* improve normalization
* concatenation of video files with ffmpeg concat filter: use remote worker (nomad)
* try to use concat demuxer before concat filter

PR #321, opened 2021-02-17
zou.utils.movie: add tests

pilou-:
**Requires #320** (since both concatenation methods are tested).

`zou.utils.movie`: add tests

For the input, a local file can be used instead of the default remote video:

    ZOU_TEST_VIDEO_URL="file:///path/to/video/file" pytest

The tests expect the video file to be with only one video stream (no audio stream) and has a resolution of 320x240.

### DIFF
--- a/tests/utils/test_movie.py
+++ b/tests/utils/test_movie.py
@@ -1,0 +1,113 @@
+import inspect
+import os
+import shutil
+import tempfile
+import unittest
+from pathlib import Path
+from urllib import request
+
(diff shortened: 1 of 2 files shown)


frankrousseau:
Hello @pilou- can you rebase this PR? I don't know why it doesn't match with #320 .

frankrousseau:
Probably because I added commits. Sorry for tat.

pilou-:
Rebased on master

PR #332, opened 2021-03-10
Change permissions for preview retrieving

LedruRollin:
**Problem**

Users with the role "Artist" can't use the gazu function [get_all_preview_files_for_task(task)](https://gazu.cg-wire.com/specs.html?highlight=get_all_previe#gazu.files.get_all_preview_files_for_task).
It seems to be a permission issue from zou, related to the fact that artists can't access the following route : api/data/preview-files

Artist :

preview2

Studio manager :

preview

**Solution**

I think the problem could be solved if the appropriate permission were added in zou here : https://github.com/cgwire/zou/blob/d33ba7393aa4435694253e17cb00a633f6c06281/zou/app/blueprints/crud/preview_file.py#L7
Since the permission function [`check_read_permissions`](https://github.com/cgwire/zou/blob/d33ba7393aa4435694253e17cb00a633f6c06281/zou/app/blueprints/crud/base.py#L111)
 was not reimplemented, I think the default behaviour was to check for admin permissions, which is maybe a little bit too much (artists can see preview files, they should be able to get them). With the proposed changes, the route would be exposed to everyone, artists and admins.
I'm still learning zou, so maybe the changes I propose are not appropriate, please let me know if there's any issue with py proposition

### DIFF
--- a/zou/app/blueprints/crud/preview_file.py
+++ b/zou/app/blueprints/crud/preview_file.py
@@ -8,6 +8,9 @@ class PreviewFilesResource(BaseModelsResource):
     def __init__(self):
         BaseModelsResource.__init__(self, PreviewFile)
 
+    def check_read_permissions(self):
+        return True
+
 
 class PreviewFileResource(BaseModelResource):
     def __init__(self):
--- a/zou/app/utils/permissions.py
+++ b/zou/app/utils/permissions.py
@@ -21,7 +21,7 @@ def has_manager_permissions():
 
 def has_admin_permissions():
     """
-    Return True if user is admin or manager.
+    Return True if user is admin.
     """
     return admin_permission.can()
 


frankrousseau:
Thank you @LedruRollin but the problem is that you open the access to everyone by returning true in the get permission function. It should return True only if the preview is part of the project to which the artist/supervisor belongs or if the user is admin.

LedruRollin:
> It should return True only if the preview is part of the project to which the artist/supervisor belongs or if the user is admin.

Thanks for quick answer @frankrousseau ! Maybe I'm misunderstanding, but isn't the behaviour you describe a job for : https://github.com/cgwire/zou/blob/d33ba7393aa4435694253e17cb00a633f6c06281/zou/app/blueprints/crud/preview_file.py#L12
which inherits from `BaseModelResource`, and decides of the behaviour of single preview files ? (from my understanding)

I want to modify the behaviour of the route api/data/preview-files (since It's the one used by the problematic gazu function), which points at all the preview files of a project (the class I want to modify inherits from BaseModel**S**Resource). So I could prevent some users from using that route (for example vendors thanks to [this function](https://github.com/cgwire/zou/blob/2a3a155247868c288f23ed449386d0680b89b675/zou/app/services/user_service.py#L350)), but I don't see how I can check if "the preview is part of the user project" (in the context of my changes, I can't even access any preview instance)

frankrousseau:
Sorry for that. The problem is a little bit more complex. Here you give access to anyone to the preview file list. You should filter the result to ensure that only preview files related to the projects the user work on can be returned. Plus you need to handle the vendor case which can access only to the tasks he is working on.

frankrousseau:
I close this due to inactivity.

PR #337, opened 2021-03-26
Less opinionated root path

ninglesby:
**Problem**
If a studio didn't want to use a `mountpoint` or `root` in their file tree and they set the values to empty strings `get_root_path()` would still return a string with separators.

**Solution**
 Check if  `root` is an empty string. If there isn't a `root` provided we can get rid of a separator. This change allows a studio to pass empty string for either mountpoint, root, or both and still get a valid path.


frankrousseau:
Great, thank you!

PR #344, opened 2021-04-26
Update api.md

keshon:
An attempt to fix dated info regarding API methods in Zou. Basically all routes should have dashes instead of underscores.

**Problem**
Documentations has outdated information about API methods.

**Solution**
Replace old routes names with dated ones (`_` replace to `-`).


frankrousseau:
Awesome, thank you!

Issue #351, opened 2021-05-28
Can't login (with ldap): KeyError: 'password'


flablog:
After our last update, user where experiencing issues while trying to connect to our instance. The instance is connected to a ldap.
The error was :
```
Traceback (most recent call last):
  File "/opt/zou/env/lib/python3.6/site-packages/zou/app/blueprints/auth/resources.py", line 187, in post
    del user["password"]
KeyError: 'password'
```
I fix it temporary, allowing users to connect, doing a try except to this line :
https://github.com/cgwire/zou/blob/812a9e4c3ff4eb190251cb7938e9137ec04150e2/zou/app/blueprints/auth/resources.py#L187

I don't know if it the better thing to do, but it fixed our issue for now.

frankrousseau:
The problem should be fixed. Can you try again?

PR #352, opened 2021-05-28
Add a route dedicated to annotation saving

frankrousseau:
**Problem**

Annotations are treated as a classic field while they are complex to handle. 

+ unrelated minor fix related to #351 

**Solution**

* Add a route dedicated to annotation saving. It will allow later to deal with complex situations like simultaneous modifications.
* Add a new event 'preview-file:annotation-update'


ldo (commented):
Just a note that, if you want to delete entries from a dictionary that might not be there, a simple way is with

    user.pop("password", None)

Also it is generally not a good idea to put in a blanket suppression of all exceptions. Make it more specific.

frankrousseau:
Thank you @ldo for the tips!

PR #357, opened 2021-06-22
Fix boolean envvar check for LDAP sync

mathbou:
**Problem**

The introduction of SSL for LDAP ( cgwire/zou@31ffd14 ) broke the boolean check for LDAP_IS_AD and LDAP_IS_AD_SIMPLE

**Solution**

Use the same code used in config.py https://github.com/cgwire/zou/blob/959e842bca02d721590c96675e33b807d886acbe/zou/app/config.py#L104-L105


frankrousseau:
Thank you!

PR #361, opened 2021-07-28
[production] add priority field to taskTypeLink

HugoDelval:
**Problem**
task types sorting was available only on the global configuration and not per-project

**Solution**
this PR adds this feature, allowing us to sort the task types per project by adding a `priority` column in the ProjectTaskTypeLink table

Linked to this PR : https://github.com/cgwire/kitsu/pull/725


zou/app/blueprints/crud/task_type.py
```diff
@@ -48,3 +54,28 @@ def post_update(self, instance_dict):
 
     def post_delete(self, instance_dict):
         tasks_service.clear_task_type_cache(instance_dict["id"])
+
+
+class TaskTypeLinksResource(Resource):
+    @jwt_required
+    def post(self):
+        data = request.json
+        task_type_link = ProjectTaskTypeLink.query.filter_by(
```
frankrousseau, line 63: Please don't use models inside resources. Use only services.

zou/app/blueprints/crud/task_type.py
```diff
@@ -48,3 +54,28 @@ def post_update(self, instance_dict):
...
+            task_type_id=data["taskTypeId"]
+        ).first()
+
+        if task_type_link is None:
+            task_type_link = ProjectTaskTypeLink(
+                project_id=data["projectId"],
+                task_type_id=data["taskTypeId"],
+            )
+
+        task_type_link.priority = int(data["priority"])
```
frankrousseau, line 74: Use `.update` method there instead of using the db object.

HugoDelval: `ProjectTaskTypeLink` doesnt inherit from the `BaseMixin`, so it doesn't have access to the `.update()` method. I can make it inherit from the `BaseMixin` if you want (it will add an `id` column too)

frankrousseau: Yes, we can inherit from the BaseMixin. I think this object is going to get more complex over time (we will need to add a status list to each item).

HugoDelval: It's done :)

PR #365, opened 2021-08-09
Change read permissions for previews

LedruRollin:
**Problem**
This PR is a follow-up to https://github.com/cgwire/zou/pull/332
The issue was about artists who could not access preview files even though they were part of the project

**Solution**
I worked on a solution based on your previous comments. I added tests to ensure proper behavior. 
Please let me know if I missed some implementation details, or if you have any remarks about the changes (code logic, coding style, etc)

### DIFF
--- a/tests/models/test_preview_file.py
+++ b/tests/models/test_preview_file.py
@@ -1,13 +1,28 @@
 from tests.base import ApiDBTestCase
 
 from zou.app.utils import fields
+from zou.app.models.task import Task
+from zou.app.services import (
+    tasks_service,
+    projects_service,
+)
 
 
 class PreviewFileTestCase(ApiDBTestCase):
     def setUp(self):
         super(PreviewFileTestCase, self).setUp()
+
+        self.generate_fixture_user_cg_artist()
+        self.user_cg_artist_id = self.user_cg_artist["id"]
+        self.generate_fixture_user_vendor()
+        self.user_vendor_id = self.user_vendor["id"]
+
+        self.generate_fixture_person()
+        self.generate_fixture_assigner()
+
+        # Create first project
         self.generate_fixture_project_status()
-        self.generate_fixture_project()
+        self.project1 = self.generate_fixture_project()
         self.generate_fixture_asset_type()
         self.generate_fixture_department()
         self.generate_fixture_task_type()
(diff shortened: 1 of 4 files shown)


LedruRollin:
(I'm trying to figure out why the two tests are failing)

frankrousseau:
Hello @LedruRollin,

Thank you for this addition. I need to review it. But what I can already see is that some requests are not efficient enough. If there are many files it may overload the database.

Don't worry about the tests. You can simply rebase your branch with master to fix them.

LedruRollin:
Thanks for the feedback @frankrousseau !
I rebased the branch and the tests are now working good, thanks :)
For the speed problem, I believe I should not do the filtering in Python, but directly in the SQL using the `query` Python variable. I'll try to implement that to solve the issue

frankrousseau:
> Thanks for the feedback @frankrousseau !
> I rebased the branch and the tests are now working good, thanks :)

Great !

> For the speed problem, I believe I should not do the filtering in Python, but directly in the SQL using the `query` Python variable. I'll try to implement that to solve the issue

Yes this is what I meant. Indexed DBs are much more efficient than Python code to filter big lists.

zou/app/blueprints/crud/preview_file.py
```diff
@@ -8,15 +13,67 @@ class PreviewFilesResource(BaseModelsResource):
...
+            current_user = persons_service.get_current_user()
+            tasks = tasks_service.get_person_tasks(current_user["id"], projects)
+            task_ids = [task["id"] for task in tasks]
+            if not str(preview_file["task_id"]) in task_ids:
+                raise permissions.PermissionDenied
+        else:
+            project_ids = [project["id"] for project in projects]
+            task = tasks_service.get_task(preview_file["task_id"])
+            if not task["project_id"] in project_ids:
+                raise permissions.PermissionDenied
```
frankrousseau, line 76: It's easier t to use :

```
user_service.check_project_access(task["project_id"])
````

File: zou/app/blueprints/crud/preview_file.py [hunk 2]
```diff
@@ -8,15 +13,67 @@ class PreviewFilesResource(BaseModelsResource):
...
+        projects = user_service.related_projects()
+        if permissions.has_manager_permissions():
+            pass
+        elif permissions.has_vendor_permissions() or \
+                permissions.has_client_permissions():
+            current_user = persons_service.get_current_user()
+            tasks = tasks_service.get_person_tasks(current_user["id"], projects)
+            task_ids = [task["id"] for task in tasks]
+            if not str(preview_file["task_id"]) in task_ids:
+                raise permissions.PermissionDenied
```
frankrousseau, line 71: It's easier t to use :

```
task = tasks_service.get_task(preview_file["task_id"])
user_service.check_working_on_entity(task["project_id"])
````

File: zou/app/blueprints/crud/preview_file.py [hunk 3]
```diff
@@ -8,15 +13,67 @@ class PreviewFilesResource(BaseModelsResource):
...
-        user_service.check_entity_access(task["entity_id"])
+        """
+        If it's a vendor, check if the user is working on the task
+        If it's an artist, check if preview file belongs to user projects
+        """
+        projects = user_service.related_projects()
+        if permissions.has_manager_permissions():
+            pass
+        elif permissions.has_vendor_permissions() or \
+                permissions.has_client_permissions():
```
frankrousseau, line 66: Client has similar permissions as artists.

LedruRollin: Thx, I made the change

tests/models/test_preview_file.py
```diff
@@ -37,6 +110,55 @@ def test_get_preview_file(self):
...
+        route1_1 = f"data/preview-files/{str(self.preview_file1_1.id)}"
+        route2_1 = f"data/preview-files/{str(self.preview_file2_1.id)}"
+        project1_id = str(self.project1.id)
+
+        self.log_in_cg_artist()
+        preview_file_artist1_1 = self.get(route1_1)
+        task_related = tasks_service.get_task(preview_file_artist1_1["task_id"])
+        self.assertEqual(task_related["project_id"], project1_id)
+        with self.assertRaises(Exception):
+            self.get(route2_1)
```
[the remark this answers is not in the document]
LedruRollin: There was even a `self.get_404(route)` ! However here I was not checking a non existing route, but rather a forbidden one (so error returned is 403 not 404). I ended up using `self.get(route, code=403)`

zou/app/blueprints/crud/preview_file.py
LedruRollin: Yeah way more simple. I ended up using [this function](https://github.com/cgwire/zou/blob/c582526a3559ac1de160c9b4d7e26d587ddab70f/zou/app/services/user_service.py#L321), which also return True if the artist/client is part of the project

zou/app/blueprints/crud/preview_file.py
LedruRollin: Here I wanted to check if the user was working on the task from the preview file, so I slightly modified the code you proposed. 
I added a function to check if the user is working on a given task (see [here](https://github.com/cgwire/zou/pull/365/files#diff-43339dcd7d47c6677253ec1e0de99bbe29c5835e07238d153359090830a98cf7R310)) ; I think the `check_working_on_entity` function could not answer this

frankrousseau:
It looks great, thank you! There is just an error with admin rights. Admins can see everything. Managers have the same read rights as artists. But don't worry, I will fix that.

LedruRollin:
Ok thanks Frank. I let the discussions unsolved, let me know if you need anything

frankrousseau:
You did 99% of the work don't worry :)

frankrousseau:
I merged it to avoid another back and forth.

PR #368, opened 2021-08-16
Add start service line to jobs documentation

Animatect:
add information missing on zou jobs page about Setting up RQ (start service after creting file, specify which is the main service file)

**Problem**
information missing on zou jobs page about Setting up RQ

**Solution**
add this documentation, (start service, specify which is the main service file)


frankrousseau:
Great, thank you a lot @Animatect!

PR #371, opened 2021-08-23
vscode conf

EvanBldy:
**Problem**
 No vscode configuration for debugging
**Solution**
 Create this configuration

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,4 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
frankrousseau, line 4: Usually, you cannot use comments in `.json` files. I think it's better if you remove them.

EvanBldy: It was generated by VSCode directly, so I let them. But yeah I can remove them.

frankrousseau:
Thx @EvanBldy

PR #380, opened 2021-09-22
Fix upgrade-db command

aditiapratama:
In latest zou, the command has changed.

Fix docs from `upgrade_db` to `upgrada-db`


frankrousseau:
Great, thank you!

PR #382, opened 2021-10-05
Various fixes

frankrousseau:
**Problem**

* Preview file route is broken
* Casting CSV import doesn't handle updates
* Sometimes users want to upload their own videos as previews and don't want normalization
* When requesting is-authenticated route, the user departments are not given. This route is use to build the kitsu context. So it's a problem to not have the user departments at that time.
* Artists can't assign tasks to themselves

**Solution**

* Fix sqlachemy usage error
* Update nb of occurrences and label on CSV import
* Allow uploading a video without running the normalization. In that case, low def and high def videos are similar.
* Add department route to user infos returned by is-authenticated route.
* Allow artists to assign themselves tasks if they belong to the task department.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -464,7 +464,7 @@ class ApiDBTestCase(ApiTestCase):
             email=u"john.did.cg.artist@gmail.com",
             role="user",
             password=auth.encrypt_password("mypassword"),
-        ).serialize()
+        ).serialize(relations=True)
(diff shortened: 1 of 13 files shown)


EvanBldy (approved):
Everything seems ok

EvanBldy:
Thanks @frankrousseau

PR #398, opened 2021-11-17
Get department from task AND task_type

Plyrolith:
\<Department> file tree variable was not working for output files, since the department id lookup only happened for tasks and not for task types.

**Solution**
Added **get_department_from_task_type** function.
For more streamlining, the existing **get_department_from_task** will now also use this function as a base, since it already looked up the task's task type.
**get_folder_from_department** now includes both functions and takes *task_type* as an additional parameter.


frankrousseau:
Great, thank you! Next time, please think about adding tests to our test suite. It we be better for further maintenance

PR #401, opened 2021-11-23
Fix normalize_movie when add empty soundtrack and improve logs

nledez:
**Problem**
Error when normalize movies without soundtrack

**Solution**
Add an error code management


frankrousseau:
Great, thank you!

PR #418, opened 2022-02-03
Improve error messages for ffmpeg exceptions

nledez:
**Problem**
When ffmpeg failed we have only a Python error.

**Solution**
Add ffmpeg output.

### DIFF
--- a/zou/utils/movie.py
+++ b/zou/utils/movie.py
@@ -14,6 +14,18 @@ EncodingParameters = namedtuple(
 )
 
 
+def print_ffmpeg_error(e):
+    print("Error:")
+    if e.stdout:
+        print("stdout:")
+        print(e.stdout.decode())
+        print("======")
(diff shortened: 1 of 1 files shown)


zou/utils/movie.py
```diff
@@ -36,9 +36,21 @@ def generate_thumbnail(movie_path):
...
     file_target_path = os.path.join(folder_path, file_target_name)
 
-    ffmpeg.input(movie_path, ss="00:00:00").output(
-        file_target_path, vframes=1
-    ).run(quiet=True)
+    try:
+        ffmpeg.input(movie_path, ss="00:00:00").output(
+            file_target_path, vframes=1
+        ).run(quiet=True)
+    except ffmpeg._run.Error as e:
```
frankrousseau, line 43: The code that handles the error is similar every time, can you make a helper function to display and raise the error?

PR #419, opened 2022-02-03
new route to get all asset tasks linked to an episode

EvanBldy:
**Problem**

- There's no route to get all asset tasks linked to an episode
- There's no route to get all attachment-files for a specific task

**Solution**

- Add a route to get all asset tasks linked to an episode
- Add a route to get all attachment-files for a specific task


frankrousseau:
Thank you @EvanBldy !

PR #422, opened 2022-02-10
various commits

EvanBldy:
**Problem**

* There's no route to get the time-spent for a specific user, task and date
* some files needs to be formatted with black

**Solution**
* add a route to get the time-spent for a specific user, task and date
* black on all py files


zou/app/services/time_spents_service.py
```diff
@@ -149,6 +149,19 @@ def get_time_spents(person_id, date):
     return fields.serialize_list(time_spents)
 
 
+def get_time_spent(person_id, task_id, date):
+    """
+    Return time spents for given person, task and date.
+    """
+    try:
+        time_spent = TimeSpent.query.filter_by(
```
frankrousseau, line 157: you should use plural for the variable name: `time_spents`

EvanBldy: I use first() of all() so it's only a "time_spent"

PR #429, opened 2022-02-24
Normalize improvements

nledez:
**Problem**
- Some error looklike `codec not currently supported in container`
- Logs are not usable

**Solution**
- Logs refactoring
- Refactoring some video stuff
- Add a transcoding if `add_empty_soundtrack` failed

### DIFF
--- a/zou/utils/movie.py
+++ b/zou/utils/movie.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 import contextlib
+import logging
 import os
 import math
 import shutil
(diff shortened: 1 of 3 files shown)


zou/utils/movie.py
```diff
@@ -84,6 +90,33 @@ def get_movie_size(movie_path):
     return (width, height)
 
 
+def ffmpeg_normalize(movie_path, task, file_target_path, fps, b, width, height):
```
frankrousseau, line 93: Can you rename it `normalize_encoding`?

zou/utils/movie.py
```diff
@@ -204,16 +191,46 @@ def add_empty_soundtrack(file_path):
...
         shutil.copyfile(tmp_file_path, file_path)
+    else:
+        logger.error(f"Err in soundtrack: {err}")
+        logger.error(f"Err code: {sp.returncode}")
+        if try_count <= 1:
+            (width, height) = get_movie_size(file_path)
+            if height % 2 == 1:
+                height = height + 1
+            stream = ffmpeg.input(file_path)
+            stream = ffmpeg.output(
```
frankrousseau, line 213: Why don't you use the function you created before (ffmpeg_normalize)?

nledez: Because I don't pass the arguments :
- `stream.audio`
- `r=fps,`
- b=b,`

This could be fixed in a future refactoring.

PR #431, opened 2022-03-05
Default timezone added to config

mlesin:
**Problem**
Previously, default time zone for users was hard coded into value "Europe/Paris"

**Solution**
Now it is moved into config, allowing administrators to change it by setting `DEFAULT_TIMEZONE` environment variable.
If not set, previous behavior is preserved.


frankrousseau:
Awesome! Thank you! I let @EvanBldy review this tomorrow.

EvanBldy:
Hello @mlesin,
It's merged!
Thanks a lot for your PR :)

PR #436, opened 2022-03-14
Allow same name of task_type for different entities, cleanup tmp after preview upload

mlesin:
**Problem**
* Previously you couldn't add two task types with the same name but for different entities (like FX for Shot and FX for Asset) through api, while database schema definitely allow this to do so.
* Some temporary files were not properly cleaned after generating previews.

**Solution**
All of the above issues have been fixed.

### DIFF
--- a/zou/app/services/comments_service.py
+++ b/zou/app/services/comments_service.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 import re
 
 from flask import current_app
@@ -241,6 +242,7 @@ def create_attachment(comment, uploaded_file):
     size = fs.get_file_size(tmp_file_path)
     attachment_file.update({"size": size})
     file_store.add_file("attachments", attachment_file_id, tmp_file_path)
+    os.remove(tmp_file_path)
     return attachment_file.present()
 
 
--- a/zou/app/services/preview_files_service.py
+++ b/zou/app/services/preview_files_service.py
@@ -96,6 +96,7 @@ def prepare_and_store_movie(
     from zou.app import app as current_app
 
     with current_app.app_context():
+        normalized_movie_low_path = None
         try:
             project = get_project_from_preview_file(preview_file_id)
         except PreviewFileNotFoundException:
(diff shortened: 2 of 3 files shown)


frankrousseau:
Hello @mlesin, thank you for your PR. But we want to keep the fact that you can't have two task types with the same name. It's a source of confusion. We didn't enforce it at the database level. Because initially Zou was aimed at importing data from SG. Can you remove this part from the PR? 
The cleanups are very useful, thank you for this!

mlesin:
Can we discuss that? 
Actually I don't see any problem in a fact that, for example, modeling can exist on both shot and asset level. We are using this approach in our studio without any issues. Thats why I did that PR, to be able to import data from our system we used before.
What kind of issues are you think may arise from that?
Users are still free not to create activities with the same name if they don't like that...

frankrousseau:
I understand that you are ok with that but on our side, we have to provide the best experience to artists and directors. The clearer, the better. It doesn't help to have similar terms for different purposes. It's ok for people dealing all day with the data but not for all the stakeholders. Collaboration is what matters the most to us, in the end it is what leads to better pictures. That's why we don't want the same term for different concepts. 
But you can use ShotAnimation or AssetAnimation if you want, it would solve the issue you mention. It will work for your import too.

mlesin:
I can't agree that a hard constraint leads to a pure collaborative experience. 
It usually leads to more inconveniences, such as the need to have unnecessarily long names as you suggested, which in the next step will lead to longer paths on disk storage (for one-to-one mapping, so the user has a better understanding of what is where).
On the other hand, without such a restriction, there is more freedom for the user to not do things they don't like (for example, you can avoid creating the same names) and only do what they really need.
Of course, you are free not to apply this PR, but it's an absolutely necessary feature on our part to be able to work in a more predictable and efficient way.

mlesin:
cleanup commits now rebased to isolate from type_task change

frankrousseau:
Thank you a lot!

PR #442, opened 2022-03-22
[search filter] allow search filters to have more characters (2000 ma…

EvanBldy:
…x, it was 200)

**Problem**
If we use search filters with a lot of filters sometimes it's not possible to save them due to their size (in DB it's only possible to save 200 chars)

**Solution**
Increase the db column (search_query) to max 2000 chars  to store bigger search filters


frankrousseau:
2000 is huge. Can you make it 500?

frankrousseau:
Thx!

PR #445, opened 2022-03-24
Add preview files revisions to edit API payload

ex5:
**Problem**
We'd like to be able to load all Edit preview files in a single request, similarly to how preview files are loaded on the Playlist page.
This would allow using Kitsu's `RawVideoPlayer` on Edit page similarly to `PlaylistPlayer`, for quicker access to task previews and revisions.

**Solution**
Add `preview_files` to `edits/with-tasks` API response.
This change is reusing some code from playlists service, moving it into another module and adjusting it to make it usable with edits.

**Additional context**
This PR is part of an overhaul of the Edit page. The goal is to make Edit page look more like a Playlist page, with main focus on the player. This was mentioned in https://github.com/cgwire/kitsu/discussions/801 as the second iteration of the front-end implementation, see the last screenshot in the proposal there.

For better idea of why this is needed, here's the updated (but yet unfinished) new version of the Edit page:

https://user-images.githubusercontent.com/176934/159978236-7adddae6-1ad8-4aaf-81ce-7c6a752d3046.mp4


ex5:
As far as I can tell, failed tests in the [latest build for this PR](https://app.travis-ci.com/github/cgwire/zou/jobs/564707270) are not failing because of this change, so it's ready for review.

ex5:
Hm, it looks like there might be a better way to achieve the same functionality (by using `/data/playlists/entities/<entityID>/preview-files` endpoint), so I'll close this for now and reopen later if this turns out to be needed after all.

frankrousseau:
Sorry for the late reply. I didn't take time to reflects properly about the proposal. And yes tests are broken. I'm working on it actually.

PR #448, opened 2022-03-30
Pass preview file ID via the room

ex5:
For having review rooms that allow switching between different preview files without having to leave room or the page.


frankrousseau:
Thank you @ex5

Issue #405, opened 2022-01-12
`alembic_version` still exist after `zou clear-db` 


nledez:
Hi, when you launch `zou clear-db` the table `alembic_version` still exists and filled with the last migration.

With this table and `version_num` defined the command `zou init-db` do noting.

To reproduce:
- Launch `zou init-db` to have a fresh DB
- In Postgres `\d` list all tables
- Launch `zou clear-db` to clean the DB
- In Postgres `\d` show `alembic_version` table that remains
- `select * from alembic_version;` show the last `version_num`
- Launch `zou init-db`

```
Creating database and tables...
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
Database and tables created.
```

frankrousseau:
You're right, it's not optimal. You have to run the `zou reset-migrations` command.

frankrousseau:
@nledez Can I Close this issue?

PR #449, opened 2022-03-30
[cli] reset-db: clear-db and stamp to first revision before, clear-db…

EvanBldy:
…: stamp to first revision before

**Problem**
Command clear-db : we don't set the db revision to first one when we clear the db, so when we do init-db it's not working (see https://github.com/cgwire/zou/issues/405)
Command reset-db is not working because it calls clear-db and we need to set the db revision to first one

**Solution**
For clear-db / reset-db : set the db revision to first one
For reset-db : remove the call to clear-db copy the code instead

### DIFF
--- a/zou/cli.py
+++ b/zou/cli.py
@@ -73,12 +73,12 @@ def clear_db():
     with app.app_context():
         import zou
 
-        directory = os.path.join(os.path.dirname(zou.__file__), "migrations")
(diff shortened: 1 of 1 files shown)


frankrousseau:
Can you stamp the database after the delection occurs? That way when a deletion fails, it doesn't change the alembic version stored in the database.

frankrousseau:
Thx!

PR #452, opened 2022-04-11
Feature: multi episode by asset

Tilix4:
**Problem**
Implementation for https://github.com/cgwire/kitsu/pull/847

**Solution**
- In `Breakdown`: Implementation using `EntityLink` and based on `get_sequence_casting` system but for episodes.
- In `Assets`: Episodes the asset is casted in are fetched by `get_assets_and_tasks` under the `episodes_names` key.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,7 +55,7 @@ install_requires =
     python-slugify==3.0.2
     boto3
 
-    pillow==7.2.0
+    pillow==8.4.0
     pytz==2020.4
     babel==2.7.0
     isoweek==1.3.3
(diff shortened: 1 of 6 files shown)


frankrousseau:
Thank you for this PR! Please can you not use the instance term? It is used for another concept. It may lead to some confusion.

Tilix4:
> Thank you for this PR! Please can you not use the instance term? It is used for another concept. It may lead to some confusion.

Sure, sorry it was dead code from an earlier step.

I'm putting this PR to `WIP` back because I've found some wrong behaviour...

frankrousseau:
The code was added to master

PR #458, opened 2022-04-28
Fix breakdown import

EvanBldy:
**Problem**
- when we upload a csv for breakdown/casting it's always needed to have an "episode" column even though it's not a tv show
- the casting stats are not correctly upgraded

**Solution**
- "episode" column not needed if it's not a tv show 
- correctly refresh casting stats 

This PR is related to this one : https://github.com/cgwire/kitsu/pull/865

### DIFF
--- a/zou/app/blueprints/source/csv/casting.py
+++ b/zou/app/blueprints/source/csv/casting.py
@@ -89,7 +89,9 @@ class CastingCsvImportResource(BaseCsvProjectImportResource):
                 link.update({"nb_occurences": occurences, "label": label})
             entity_id = str(entity.id)
             if shots_service.is_shot(entity.serialize()):
-                breakdown_service.refresh_shot_casting_stats(entity.serialize())
+                breakdown_service.refresh_shot_casting_stats(
+                    entity.serialize()
+                )
                 events.emit(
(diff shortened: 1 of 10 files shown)


zou/app/blueprints/source/csv/casting.py
```diff
@@ -67,10 +78,29 @@ def import_row(self, row, project_id):
...
+            entity = entities_service.get_entity_raw(target_id)
             link = breakdown_service.get_entity_link_raw(target_id, asset_id)
             if link is None:
                 breakdown_service.create_casting_link(
                     target_id, asset_id, occurences, label
                 )
+                entity.update({"nb_entities_out": entity.nb_entities_out + 1})
             else:
                 link.update({"nb_occurences": occurences, "label": label})
+            breakdown_service.refresh_shot_casting_stats(entity.serialize())
```
frankrousseau, line 90: This line should be after the test condition verifying the shot type.

frankrousseau (commented):
Could you add tests for this bugfix?

zou/app/blueprints/source/csv/casting.py
EvanBldy: It's done

PR #476, opened 2022-05-27
refactoring for import csv

EvanBldy:
**Problem**
- for assets/shots/edits imports for example description is deleted when we do an update and the column is missing 
- the code sometimes is hard to understand (for tasks update of assets/shots/edits for example)
**Solution**
- assets/shots/edits/persons refactoring of the code + check all behaviours to stop deleting informations

### DIFF
--- a/zou/app/blueprints/source/csv/assets.py
+++ b/zou/app/blueprints/source/csv/assets.py
@@ -1,4 +1,7 @@
-from zou.app.blueprints.source.csv.base import BaseCsvProjectImportResource
+from zou.app.blueprints.source.csv.base import (
+    BaseCsvProjectImportResource,
+    ImportRowException,
+)
 from zou.app.models.project import ProjectTaskTypeLink
 from zou.app.models.task_type import TaskType
 
@@ -9,6 +12,7 @@ from zou.app.services.tasks_service import (
     create_tasks,
     get_tasks_for_asset,
     get_task_statuses,
+    get_task_type,
 )
 from zou.app.services.comments_service import create_comment
 from zou.app.services.persons_service import get_current_user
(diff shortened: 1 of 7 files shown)


zou/app/blueprints/source/csv/assets.py
```diff
@@ -52,14 +56,83 @@ def prepare_import(self, project_id):
...
+        for task_type in self.task_types_in_project_for_assets:
+            task_status_name = row.get(task_type.name, None)
+            task_status_id = None
+            if task_status_name is not None:
+                for status_id, status_names in self.task_statuses.items():
+                    if task_status_name in status_names:
+                        task_status_id = status_id
+                        break
+                if task_status_id is None:
+                    raise TaskStatusNotFoundException(
```
frankrousseau, line 70: Be careful, raising exceptions like this generates a 404 error.

zou/app/blueprints/source/csv/assets.py
```diff
@@ -68,6 +141,10 @@ def import_row(self, row, project_id):
                     "id"
                 ]
             episode_id = self.episodes.get(episode_name, None)
+        elif episode_name is not None:
+            raise ImportRowException(
```
frankrousseau, line 145: Same here. You can't raise exception in the midlle of the code without handling them somewhere.

frankrousseau:
Tests are broken. Could you fix them and add some more?

zou/app/blueprints/source/csv/assets.py
EvanBldy: Effectively, I have to use ImportRowException

zou/app/blueprints/source/csv/assets.py
EvanBldy: I use RowException now that is handled in the main class

PR #482, opened 2022-06-09
[qa] fix migration : unexisting table asset_types

EvanBldy:
**Problem**
When we do an python zou/cli.py upgrade-db we got a problem with an unexisting table asset_types

**Solution**
Remove unexisting table


frankrousseau:
Thank you!

PR #487, opened 2022-06-20
AD users may have complex flags combination in userAccountControl

mathbou:
**Problem**

The userAccountControl flag may contain a complex combination of flag. Some users end up disabled despite the absence of the ACCOUNT_DISABLED flag.

**Solution**

Check the absence of the ACCOUNT_DISABLED flag (2) instead of predefined flags combination


frankrousseau:
Thank you @mathbou for your PR. But the code is not explicit and I'm not sure to understand your explanation. Could you make things clearer?

PR #488, opened 2022-06-21
various commits

EvanBldy:
**Problem**
- In RELEASE.md environment variable release_number is declared as release
- For day-offs user can't assign themselves

**Solution**
- In RELEASE.md declare release_number instead of release
- Fix day-offs permissions


frankrousseau:
Thx!

Issue #464, opened 2022-05-11
Missing project_id field from events


Tilix4:
`project_id` field is missing from `episode:delete` and `sequence:delete`, when it's provided by `asset:delete`.

PR #499, opened 2022-07-07
Add project_id data field for episode sequence delete events

kaamaurice:
## Problem
When adding **gazu** event listeners `episode:delete` and `sequence:delete`,  data don't provide `project_id` field as `asset:delete` or `shot:delete`

### Solution
- Just adding project_id with `events.emit` in the data argument.

### Additional note
Not sure about this **PR** because when `asset:delete` events are emitted we have `project_id` in data with **gazu** listeners but seems not provided in **zou** `assets_service.py`

https://github.com/cgwire/zou/blob/ad7895b5c3e005302310ac1f553ffc7d77c6dc18/zou/app/services/assets_service.py#L669

Resolve: #464

### DIFF
--- a/zou/app/services/deletion_service.py
+++ b/zou/app/services/deletion_service.py
@@ -381,7 +381,7 @@ def remove_episode(episode_id, force=False):
         episode.delete()
         events.emit(
             "episode:delete",
-            {"episode_id": episode_id},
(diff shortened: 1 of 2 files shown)


kaamaurice:
Seems solved since https://github.com/cgwire/zou/pull/484

PR #510, opened 2022-08-10
[qa] only allow to create task for task types corresponding to their …

EvanBldy:
…entities type

**Problem**
We are allowed to add tasks to an entity, which their task types are not corresponding to their entity types

**Solution**
Disable this possibility


zou/app/blueprints/crud/task.py
```diff
@@ -46,6 +50,19 @@ def post(self):
...
+            if task_type["for_entity"] == "Asset":
+                if not assets_service.is_asset_dict(entity):
+                    raise WrongTaskTypeForEntityException
+            elif (
+                entities_service.get_temporal_entity_type_by_name(
+                    task_type["for_entity"]
+                )["id"]
+                != entity["entity_type_id"]
+            ):
+                raise WrongTaskTypeForEntityException
```
frankrousseau, line 64: You should handle this error the same way other errors are managed: 
https://github.com/cgwire/zou/blob/master/zou/app/__init__.py#L81-L83

PR #514, opened 2022-08-23
Fix the parameters of files and update the source of the openAPI file

NehmatH:
**Problem**
- There is no parameters for file input in most of the endpoints.
- The openAPI file used in the Github Actions comes from a site not controlled by CGWire.
- The models were written with no specific naming conventions.

**Solution**

- Add file parameters in all endpoints.
- Update the openAPI file based on a testing instance called kitsu-staging that is updated and modified by CGWire.
- Use CamelCase as a naming convention for all models.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,9 +30,10 @@ jobs:
     - name: Update openapi.json
       run: |
         cd docs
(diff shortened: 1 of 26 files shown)


nledez (approved):
LGFM (for the workflow part)

EvanBldy:
Thanks @NehmatH :)

PR #515, opened 2022-08-23
Add 'filter by tags' feature and modify css

NehmatH:
**Problem**
- No filter for modules

**Solution**
- Add 'filter by tags' feature and make it case insensitive


EvanBldy:
Thanks @NehmatH

Issue #513, opened 2022-08-22
cache files won't get written to tmp from s3 storage


cbsrobot:
after upgrading zou from 0.13.x to 0.15.1 I get the following error:

```
Traceback (most recent call last): File "/opt/zou/zouenv/lib/python3.8/site-packages/flask/app.py", line 1949, in full_dispatch_request
 rv = self.dispatch_request()
 File "/opt/zou/zouenv/lib/python3.8/site-packages/flask/app.py", line 1935, in dispatch_request
 return self.view_functions[rule.endpoint](**req.view_args)
 File "/opt/zou/zouenv/lib/python3.8/site-packages/flask_restful/__init__.py", line 468, in wrapper
 resp = resource(*args, **kwargs)
 File "/opt/zou/zouenv/lib…/zouenv/lib/python3.8/site-packages/zou/app/blueprints/previews/resources.py", line 105, in send_picture_file
 return send_storage_file(
 File "/opt/zou/zouenv/lib/python3.8/site-packages/zou/app/blueprints/previews/resources.py", line 129, in send_storage_file
 file_path = fs.get_file_path_and_file(
 File "/opt/zou/zouenv/lib/python3.8/site-packages/zou/app/utils/fs.py", line 43, in get_file_path_and_file
 tmp_file.write(chunk)
TypeError: a bytes-like object is required, not 'int'
```

and the cache files are not written to the tmp dir.

I traced it back to https://github.com/cgwire/zou/commit/7e9cf97483151ef95a9a11dfdc15d207854a0cb5

reverting the hunk `for chunk in io.BytesIO(read_stream):`  seems to solve the issue

EvanBldy:
Hi @cbsrobot, 
Thanks for the report !
It should be fixed in this PR : https://github.com/cgwire/zou/pull/520.
Can you close the ticket if it's good for you ?

Issue #491, opened 2022-06-24
Zou commands raise error `Permission denied: 'indexes'` on each call


nledez:
If you aren't in `/opt/zou` folder and not with user `zou` commands (`init-db`, `upgrade-db`, `create-admin` and `init-data` for now).

frankrousseau:
You have to set the `INDEXES_FOLDER` variable to the folder you want to avoid this problem.

nledez:
Nope. But it’s not in documentation.
(in old doc - maybe use by other users):
’DB_PASSWORD=yourdbpassword zou upgrade-db ’
It’s failed with ’Permission denied’ error. Or it creates a ’indexes’ folder on the current path (wrong one).

gstorni:
Hello, I got the same issue trying to backup the database
<img width="844" alt="Screen Shot 2022-07-07 at 22 43 50" src="https://user-images.githubusercontent.com/3090235/177899965-aa80c5ca-3798-4a01-b9b7-a71fb5bfc5d0.png">

I found this waklkaround. 
https://stackoverflow.com/questions/67863292/permissionerror-errno-13-permission-denied-on-ubuntu-trying-to-run-selenium

"sudo chmod -R 777 /path/to/gecko"

and it worked.
<img width="1004" alt="Screen Shot 2022-07-07 at 22 51 14" src="https://user-images.githubusercontent.com/3090235/177900609-0da6f8c4-168f-49a0-a4a2-d9ddfacf1a53.png">

It happend the same in a Google Cloud Ubuntu Instance with Postgresql localhost and locally in a Multipass Ubuntu Instance with the database in another instance.

Hope it helps!

frankrousseau:
Please don't do that. It gives access rights to everyone.  A `chown` for the `zou` user should be enough.

gstorni:
Thank you Frank, Ill try that.

EvanBldy:
Hi @nledez,
It should be fixed in this PR https://github.com/cgwire/zou/pull/520

PR #520, opened 2022-08-31
various commit

EvanBldy:
**Problem**
- pre-commit for black not working
- it's not allowed to publish .webm files
- search indexes folder are created when we launch a command via zou (it can set wrong rights) : https://github.com/cgwire/zou/issues/491
- environment variables are handled in multiple files of the code we need to refactor that
- Amazon S3 are not working : https://github.com/cgwire/zou/issues/513

**Solution**
- update pre-commit and black to fix
- allow to publish .webm files
- create search indexes folder only when needed (and add a command to create it) + refactoring
- refactoring for environment variables
- Fix Amazon S3 + read files with chunks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ Vagrantfile
 
 /previews/
 /thumbnails/
+/indexes/
 release.sh
 
 tmp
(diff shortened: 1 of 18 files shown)


EvanBldy:
> d the `[qa] also read chunks for amazon s3 bucket` part and it works as expected. Thanks a lot.

Thanks for your review @cbsrobot :)

PR #521, opened 2022-09-06
[qa] fix tests/models/test_person.py

EvanBldy:
**Problem**
Test for tests/models/test_person.py : test_delete_person are randomly broken because sometimes it tries to delete the user making the request.

**Solution**
Sort persons by first_name and last_name for route data/persons


frankrousseau:
It's ok for me. Let's wait for Travis tests to finish.

PR #525, opened 2022-09-23
[permissions] vendor don't see every metadata descriptors

EvanBldy:
**Problem**
- vendors see every metadata descriptors and every data information

**Solution**
- vendor don't see every metadata descriptors, they see only metadata descriptors with no departments or metadata descriptors corresponding to their departments, the same for data in entity

### DIFF
--- a/zou/app/blueprints/assets/resources.py
+++ b/zou/app/blueprints/assets/resources.py
@@ -139,9 +139,10 @@ class AssetsAndTasksResource(Resource):
             criterions["assigned_to"] = persons_service.get_current_user()[
                 "id"
             ]
-            criterions[
-                "vendor_departments"
-            ] = persons_service.get_current_user(relations=True)["departments"]
+            criterions["vendor_departments"] = [
+                str(department.id)
+                for department in persons_service.get_current_user_raw().departments
+            ]
         return assets_service.get_assets_and_tasks(criterions, page)
 
 
--- a/zou/app/blueprints/assets/resources.py
+++ b/zou/app/blueprints/assets/resources.py
@@ -139,6 +139,9 @@ class AssetsAndTasksResource(Resource):
             criterions["assigned_to"] = persons_service.get_current_user()[
                 "id"
             ]
+            criterions[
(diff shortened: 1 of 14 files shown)


frankrousseau:
It breaks, many tests. It requires some fixes.

EvanBldy:
> It breaks, many tests. It requires some fixes.

It should be fixed in the next tests :)

zou/app/blueprints/assets/resources.py
```diff
@@ -139,6 +139,9 @@ def get(self):
             criterions["assigned_to"] = persons_service.get_current_user()[
                 "id"
             ]
+            criterions[
```
frankrousseau, line 142: It should not be added if the user is not a vendor.

zou/app/services/assets_service.py
```diff
@@ -258,6 +259,15 @@ def get_assets_and_tasks(criterions={}, page=1, with_episode_ids=False):
                 str(link.entity_in_id)
             )
 
+    query_result = tasks_query.all()
+
+    if "vendor_departments" in criterions:
+        not_allowed_descriptors_field_names = entities_service.get_not_allowed_descriptors_fields_for_entity_type_and_departments(
```
frankrousseau, line 265: * The function name is too long
* The line limit width is 80 chars

zou/app/blueprints/assets/resources.py
EvanBldy: It's the case there's permissions.has_vendor_permissions() line 138

zou/app/services/assets_service.py
EvanBldy: It's fixed

frankrousseau:
It seems good. But it requires to add some tests to ensure the expected behaviour is respected.

EvanBldy:
> It seems good. But it requires to add some tests to ensure the expected behaviour is respected.

I will add them next week :)

EvanBldy:
@frankrousseau the tests are added, I think it's ready to be merged :)

PR #533, opened 2022-10-13
Authentication improvements + some small improvements

EvanBldy:
**Problem**
- DEBUG_PORT env variable is not in zou/app/config.py
- Route /data/persons/<person_id>/desktop-login-logs is not properly documented for Swagger.
- config.USER_LIMIT is not used when we do an LDAP sync. 
- We use "default" as default password when creating user.
- Send emails part for playlists need to be refactored to use zou.app.utils.emails.send_email.
- There's no route for admins to change passwords of other users. 
- Add a flag to person "is_generated_from_ldap" to know if the person is imported from an LDAP.
- Users are not notified when their password is changed.
- We need a security when a lot of logins fails for the same user.

**Solution**
- DEBUG_PORT env variable is now in zou/app/config.py
- Route /data/persons/<person_id>/desktop-login-logs is now properly documented for Swagger.
- config.USER_LIMIT is now used when we do an LDAP sync and stopping when the limit is reached.
- Stop using "default" password everywhere when we create a new user, create a token and send it with an URL to change the password instead. 
- Send emails part for playlists need tis refactored to use zou.app.utils.emails.send_email.
- Add a route for admins to change passwords of other users : "/actions/persons/<person_id>/change-password"
- Add a flag to person "is_generated_from_ldap" to know if the person is imported from an LDAP. Remove all "isLdap" because its not needed anymore.
- Users are notified by mail when their password is changed.
- Add a security when a lot of logins fails for the same user : force to wait 1 minute when 5 logins fails for same user.


tests/auth/test_auth_route.py
```diff
@@ -87,7 +87,7 @@ def test_login_wrong_credentials(self):
             "email": self.person_dict["email"],
             "password": "wrongpassword",
         }
-        result = self.post("auth/login", credentials, 400)
```
frankrousseau, line 90: Wrong credentials should return a 400 error, it's a problem with the param not the access to the route.

zou/migrations/versions/82ee682204ab_add_is_generated_from_ldap_column_to_.py
```diff
@@ -0,0 +1,117 @@
...
+        "person",
+        sa.Column("is_generated_from_ldap", sa.Boolean(), default=False),
+    )
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    persons = session.query(Person).all()
+    for person in persons:
+        if person.password == b"default":
+            person.password = None
+            person.is_generated_from_ldap = True
```
frankrousseau, line 102: By default, it should be set with False.

zou/cli.py
```diff
@@ -174,18 +174,33 @@ def init_data():
...
+@click.option("--password", required=True, default=None)
+def change_password(email, password):
+    """
+    Change the password of given user.
+    """
+    try:
+        auth.validate_password(password)
+        password = auth.encrypt_password(password)
+        persons_service.update_password(email, password)
+        print("Password changed for %s" % email)
```
frankrousseau, line 186: "The password is too short".

zou/app/utils/emails.py
```diff
@@ -1,21 +1,52 @@
...
+        self.text = StringIO()
+
+    def handle_data(self, d):
+        self.text.write(d)
+
+    def get_data(self):
+        return self.text.getvalue()
+
+
+def strip_html_tags(html):
```
frankrousseau, line 49: When do you use this function?

zou/app/blueprints/persons/resources.py
```diff
@@ -75,16 +77,9 @@ def post(self):
                 "limit": config.USER_LIMIT,
             }, 400
         else:
-            person = persons_service.create_person(
```
frankrousseau, line 78: You should keep this., it ensures that data is properly used.

zou/app/blueprints/persons/resources.py
```diff
@@ -1052,3 +1047,113 @@ def delete(self, person_id, department_id):
...
+            current_user = persons_service.get_current_user()
+            if (
+                persons_service.is_admin(person)
+                and person["id"] != current_user["id"]
+            ):
+                raise permissions.PermissionDenied
+            auth.validate_password(password, password_2)
+            password = auth.encrypt_password(password)
+            persons_service.update_password(person["email"], password)
+            current_app.logger.info(
```
frankrousseau, line 1105: Should we use the warn level instead?

zou/app/services/persons_service.py
```diff
@@ -334,66 +339,43 @@ def invite_person(person_id):
     Send an invitation email to given person (a mail telling him/her how to
     connect on Kitsu).
     """
+    person = get_person(person_id)
     organisation = get_organisation()
-    person = get_person_raw(person_id)
+    token = str(uuid.uuid4())
```
frankrousseau, line 344: The token length may be too small for brute force attacks. It's recommended to use at least 64 chars.

zou/app/services/auth_service.py
```diff
@@ -27,153 +31,136 @@ def check_auth(app, email, password):
...
+            raise NoAuthStrategyConfigured()
+    except WrongPasswordException:
+        persons_service.update_login_failed_attemps_person(
+            person["id"], login_failed_attemps + 1, datetime.now()
+        )
         raise WrongPasswordException()
 
+    persons_service.update_login_failed_attemps_person(person["id"], 0)
 
-def no_password_auth_strategy(email):
```
frankrousseau, line 82: Please do not remove the no password strategy.

zou/app/services/auth_service.py
```diff
@@ -27,153 +31,136 @@ def check_auth(app, email, password):
...
-        if app is not None:
-            app.logger.error("Wrong password for: %s" % person["full_name"])
+            raise NoAuthStrategyConfigured()
+    except WrongPasswordException:
+        persons_service.update_login_failed_attemps_person(
+            person["id"], login_failed_attemps + 1, datetime.now()
+        )
         raise WrongPasswordException()
 
+    persons_service.update_login_failed_attemps_person(person["id"], 0)
```
frankrousseau, line 74: Can you find a way to not generate requests to database each time a login succeeds?

tests/auth/test_auth_route.py
EvanBldy: An RFC (7235) says that it should return 401 as status code. 
https://stackoverflow.com/questions/32752578/whats-the-appropriate-http-status-code-to-return-if-a-user-tries-logging-in-wit

zou/migrations/versions/82ee682204ab_add_is_generated_from_ldap_column_to_.py
EvanBldy: The `b"default"` was the only way to guess who are from an ldap or not.
The others were encrypted password for "default".

zou/cli.py
EvanBldy: It's fixed

zou/app/utils/emails.py
EvanBldy: In send_email function in the same file. It permit to have a body without html (when only html is given as parameter).

zou/app/blueprints/persons/resources.py
EvanBldy: It's fixed

zou/app/blueprints/persons/resources.py
EvanBldy: Yep it's a good idea. It's done

zou/app/services/auth_service.py
EvanBldy: I have re-added a function for that but it just return the person given in parameter.

zou/app/services/persons_service.py
EvanBldy: It's done!

zou/app/services/auth_service.py
EvanBldy: Yeah, it's done :)

zou/migrations/versions/82ee682204ab_add_is_generated_from_ldap_column_to_.py
frankrousseau: You should add a comment for that.

Issue #540, opened 2022-11-04
LDAP: use of a unpopulated attribute for authentication, impossibility to connect.


autoiue:
In the file [auth_service.py line 121](https://github.com/cgwire/zou/blob/e8c9cd3069a833fa36ea6822b390c03365f21446/zou/app/services/auth_service.py#L121), the attribute `full_name `of the `person` object does not exist.

This leads to a bad value being used for authentificatino against the LDAP server.

Replacing `full_name` by `desktop_login` seems to work.

See pull request #541

EvanBldy:
Hi @procsynth,
As I said in your PR full_name it's a calculated value from the first and last name. See here : https://github.com/cgwire/zou/blob/ece10754ded8617f63a206fb538d2ace9312f811/zou/app/models/person.py#L73

Are you trying to connect with Administrator account ? 
He doesn't have a first name (givenName) and last_name (sn) so it can't work. I will fix that.

EvanBldy:
It was fixed for Administrator in this PR : https://github.com/cgwire/zou/pull/547

frankrousseau:
@procsynth could you confirm it was fixed?

PR #541, opened 2022-11-04
Fix issue #540

procsynth:
Replaces unpopulated attribute 'full_name' with 'desktop_login'

**Problem**
`full_name` does not exist in the `person` object, this leads to bad data being sent to the LDAP server

**Solution**
Change `full_name` to `desktop_login`


frankrousseau:
@EvanBldy can you review this? I'm not sure that the `desktop_login` is the right field for this.

frankrousseau:
Thank you @procsynth for the PR.

EvanBldy:
Hi, I will check that next week. I have to create a VM with MS AD to check that.

EvanBldy:
Hi @procsynth, I tested the code and it does not work for me. Have you tried your code ? It works on your AD ?

PR #547, opened 2022-11-28
refactoring LDAP + new route for getting time-spents of a task

EvanBldy:
**Problem**
- LDAP code can be refactored.
- LDAP accounts that are not present anymore have to be disabled.
- There's no route to get time spents of a specific task without specifying the date. 
- Black formater.

**Solution**
- refactoring of LDAP code to be simpler.
- LDAP accounts that are not present anymore will be disabled.
- Add new route to get time-spents of a specific task without specifying the date : /actions/tasks/<task_id>/time-spents
- Black formater.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ install_requires =
     flask_fs==0.6.1
     werkzeug==0.15.5
     redis==4.1.4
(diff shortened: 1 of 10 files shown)


frankrousseau:
Can you add tests for this?

PR #548, opened 2022-11-28
Add Sentry support

nledez:
**Problem**
- Errors are lost in logs
- Can't understand how to reproduce error with only the stack trace

**Solution**
Add the support of Sentry with https://github.com/getsentry/sentry-python


frankrousseau:
Can you run the `black` formatter on your code?

frankrousseau:
Thank you @nledez

PR #552, opened 2022-12-04
Fix booleans from environment variables

jasperges:
**Problem**
Using `strtobool()` means the default value gotten from `os.getenv()` must be a string that can be converted to a boolean. In the current implementation it is `True`/`False` and `strtobool()` raises an error.

**Solution**
Change the default values to `"false"`/`"true"`.

### DIFF
--- a/zou/app/config.py
+++ b/zou/app/config.py
@@ -7,7 +7,7 @@ from zou.app.utils.string import strtobool
 
 PROPAGATE_EXCEPTIONS = True
 RESTFUL_JSON = {"ensure_ascii": False}
-DEBUG = strtobool(os.getenv("DEBUG", False))
+DEBUG = strtobool(os.getenv("DEBUG", "false"))
 DEBUG_PORT = int(os.getenv("DEBUG_PORT", 5000))
 
 APP_NAME = "Zou"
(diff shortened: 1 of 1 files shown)


EvanBldy:
Hi @jasperges,
Thanks for the report and your PR,
But I can't figure how do you get an error because if a boolean is given as argument to strtobool it will return that boolean.
See here : https://github.com/cgwire/zou/blob/6448293bb3b53e5caa745c057e3986d0fd5544da/zou/app/utils/string.py#L2
So can you give the stacktrace ?

jasperges:
You are totally right. I jumped to conclusions before properly investigating my issue. The problem was that `MAIL_DEBUG` was set, but as an empty string. Setting it to `"false"` fixed my issue. :sweat_smile: 

You could consider updating the assignments to something like `MAIL_DEBUG = strtobool(os.getenv("MAIL_DEBUG") or False)` to also handle empty environment variables like I had. Although you can argue that people should be responsible for setting the environment variables properly. :innocent: 

Anyway, I'm closing this. Sorry, for wasting your time.

EvanBldy:
No problem @jasperges :)

I think for now people should be responsible for setting their environment variables properly, they will have an error if it's not the case. 

Have a good day

PR #567, opened 2023-01-04
[WIP] Up dependencies versions

frankrousseau:
WIP:

- Need to up socketio version in Gazu
- Need to up Gazu version after that
- Need to split the commit

**Problem**

Many dependencies are outdated, Flask included.

**Solution**

Up library versions. Store user id in JWT token instead of email.


frankrousseau:
This work is now handled in the Evan's PR.

PR #571, opened 2023-01-18
Various fixes

EvanBldy:
**Problem**
- subjects of mails can be refactored 
- when export csv of shots there's no assignations like in Kitsu

**Solution**
- refactoring for subjects of mails
- when exporting csv of shots directly via Zou add assignations columns


frankrousseau:
Can you add some tests with this code?

PR #574, opened 2023-01-23
Update troubleshooting.md

LarsCawley:
**Problem**

an update from a much earlier version can break the database

**Solution**

provide advice on how to update safely and a script to help


LarsCawley:
You're welcome!
Unfortunately I see a problem with the way the text is working on the website.  I need to escape my use example chevrons.  A beginner mistake!  New commit coming...

PR #576, opened 2023-01-24
Update troubleshooting.md

LarsCawley:
Escaping chevrons to fix unexpected Markdown behaviour

**Problem**
chevrons used for name/variable example caused problems with Markdown display

**Solution**
added escape \ to correct the problem


EvanBldy:
Thanks @LarsCawley

PR #577, opened 2023-01-25
update ldap query, do not import computers anymore

mdk-man:
**Problem**
LDAP query returns user **and** computer accounts

**Solution**
Add a filter to prevent computer accounts from being imported


frankrousseau:
Thank you!

PR #578, opened 2023-01-25
Update troubleshooting.md

LarsCawley:
Second attempt at escaping <, this time using `&lt;`

**Problem**
escaping the < with a backslash didn't work for the HTML on the website

**Solution**
using `&lt;` instead


LarsCawley:
Finally it works!  Thanks for your patience guys.

PR #581, opened 2023-02-08
Use value coming from configuration for redis connection

robin-vagner:
**Problem**
`get_redis_url()` did not use the database index from the configuration.

**Solution**
We use the database index from the configuration.


frankrousseau:
Thank you!

PR #582, opened 2023-02-08
Fix get_full_shots to get relations.

robin-vagner:
**Problem**
`get_full_shot` did not return the shot with its relations.

**Solution**
Update the shot with its relations.

**Notes**
I left the commit as it was initially created, but the code does not feel really straightforward, is there a simpler way to do this ?


frankrousseau:
Thank you for this improvement. The way it's coded is ok. It may be less performant but at least the code is simpler. And the response is cached, so performance is not that critical.

PR #583, opened 2023-02-08
Fix admin creation if user already exists

robin-vagner:
**Problem**
When we create a local database, we call `create_admin` from the command-line to make the current user the administrator.
The problem is that when the user already exists in the database, this command fails.

**Solution**
We make `create_admin` handle the case where the user already exists : in this case, we only change their role to `admin`.

An alternative solution could be to add a `make_admin` command-line action, if you would rather not modify `create_admin`.

### DIFF
--- a/zou/cli.py
+++ b/zou/cli.py
@@ -136,26 +136,27 @@ def create_admin(email, password):
     Create an admin user to allow usage of the API when database is empty.
     """
     try:
-        auth.validate_password(password)
-        # Allow "admin@example.com" to be invalid.
-        if email != "admin@example.com":
-            auth.validate_email(email)
-        password = auth.encrypt_password(password)
-        persons_service.create_person(
-            email, password, "Super", "Admin", role="admin"
-        )
-        print("Admin successfully created.")
-
-    except IntegrityError:
-        print("User already exists for this email.")
-        sys.exit(1)
-
-    except auth.PasswordTooShortException:
-        print("Password is too short.")
-        sys.exit(1)
-    except auth.EmailNotValidException:
-        print("Email is not valid.")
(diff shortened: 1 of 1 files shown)


zou/cli.py
```diff
@@ -136,26 +136,30 @@ def create_admin(email, password):
     Create an admin user to allow usage of the API when database is empty.
     """
     try:
-        auth.validate_password(password)
```
frankrousseau, line 139: Why did you removed the password validation?

robin-vagner: Ah, that is a mistake from applying the patch file, my bad.

I fixed it.

zou/cli.py
```diff
@@ -136,26 +136,31 @@ def create_admin(email, password):
...
-
-    except auth.PasswordTooShortException:
-        print("Password is too short.")
-        sys.exit(1)
-    except auth.EmailNotValidException:
-        print("Email is not valid.")
-        sys.exit(1)
+        person = persons_service.get_person_by_email(email)
+        if person["role"] != "admin":
+            persons_service.update_person(person["id"], {"role": "admin"})
```
frankrousseau, line 141: Can you write a message in the console telling that the user was upgraded?

robin-vagner: Done.
I realized when re-reading the code that there was still a `except IntegrityError` that might become obsolete with this modification.
Do you know if there some other reason that this exception could be raised or should I just remove it ?

frankrousseau: I think it is raised by SQL alchemy when the entry is already present. So yes, you can remove it!

robin-vagner: Alright, it's done.

PR #586, opened 2023-02-08
Get base entity type name

robin-vagner:
**Problem**
`get_base_entity_type_name` did not handle scenes.

**Solution**
We added a case for scenes.

### DIFF
--- a/zou/app/services/shots_service.py
+++ b/zou/app/services/shots_service.py
@@ -1068,6 +1068,9 @@ def get_base_entity_type_name(entity_dict):
         type_name = "Sequence"
     elif is_episode(entity_dict):
         type_name = "Episode"
+    elif is_scene(entity_dict):
+        type_name = "Scene"
+
     return type_name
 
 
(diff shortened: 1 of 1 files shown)


zou/app/services/shots_service.py
```diff
@@ -1060,13 +1059,17 @@ def get_shot_versions(shot_id):
 
 
 def get_base_entity_type_name(entity_dict):
-    type_name = "Asset"
```
frankrousseau, line 1063: Can you let the variable as declared first?

zou/app/services/shots_service.py
```diff
@@ -1060,13 +1059,17 @@ def get_shot_versions(shot_id):
...
     if is_shot(entity_dict):
         type_name = "Shot"
+    elif is_scene(entity_dict):
```
frankrousseau, line 1064: Can you put the scene tests at then end of the condition list? It occurs less often than the two others. Kitsu users don't have scenes.

zou/app/services/shots_service.py
robin-vagner: Yes, it's done.

zou/app/services/shots_service.py
robin-vagner: Makes sense, I fixed it.

PR #587, opened 2023-02-08
LDAP: update query for non-AD servers

mdk-man:
We can only be sure that the LDAP server has a computer type object if it's an Active Directory server

**Problem**

According to [Plyrolith](https://github.com/Plyrolith) the previous fix breaks sync with non-AD LDAP (like openLDAP)

**Solution**

Avoid computer type accounts only if we are connected to an Active Directory server


EvanBldy:
Thanks for your PR @mdk-man

PR #589, opened 2023-02-09
Add limit parameter to requests in zou to specify the number of entries per page

robin-vagner:
**Problem**
We wanted to be able to specify the number of entries that we wanted per page from the client, so as to be more flexible.

**Solution**
We added a `limit` parameter, that can be used like so : `gazu.client.fetch_all("working-files", params={"page": 2, "limit": 10})`


frankrousseau:
Thank you!

PR #590, opened 2023-02-09
Fix relations argument evaluated with wrong type

robin-vagner:
**Problem**
The `relation` argument is not automatically converted to boolean, which causes problems sometimes.

**Solution**
We compare the lowercase string to `"true"` instead, to ensure it works in all cases, such as when the user passed `params={"relations": True}` (with `True` as a boolean) to the client function.


frankrousseau:
Thank you!

PR #591, opened 2023-02-09
Use name arg in LastInstanceOutputFiles and LastEntityOutputFiles resources

robin-vagner:
**Problem**
We wanted to be able to filter the last output files by name.

**Solution**
We added a `name` argument to the resources' `get` methods.


frankrousseau:
Ok, thank you

PR #592, opened 2023-02-09
Add route to query all timespents for a user and a month

robin-vagner:
**Problem**
We needed a route to get the full list of timespents for a given person and month, and the existing route `"/data/persons/<person_id>/time-spents/month/<year>/<month>"` gave only an aggregated result (the sum of time spent for each task for the month + user).

**Solution**
We added a route to get all the timespents rather than an aggregated list.

**Note**
This is the kind of thing that would stay on our side with the plugins system you are planning on implementing, so if you feel no one else benefits from this route, we can absolutely keep this on our side while waiting for the plugins system.


frankrousseau:
Great, thank you! It's ok to include it, it doesn't break any business logic and follows our goal of providing simple and efficient routes adapted to animation/vfx industry.

PR #594, opened 2023-02-13
Change outputfile modification permissions

robin-vagner:
**Problem**
Output files update permissions were too restrictive for output files from asset instances.
The user had to be assigned to a task on the instance's entity, which was rarely the case since users are usually assigned to the temporal entities in which the instance is instantiated, rather than the entity of the instance itself.

Exemple:
For an instance of an asset `Rabbit`, instantiated in a shot `sh040`, the user that exports an output for this instance is usually assigned to the shot rather than the instance.

**Solution**
We changed the check to allow the modification if the user is assigned to the temporal entity of the output file (if there is one).

### DIFF
--- a/zou/app/blueprints/crud/output_file.py
+++ b/zou/app/blueprints/crud/output_file.py
@@ -42,5 +42,5 @@ class OutputFileResource(BaseModelResource):
             return True
         else:
             return user_service.check_working_on_entity(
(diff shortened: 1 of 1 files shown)


frankrousseau:
Thank you @robin-vagner and @pcharmoille!

PR #598, opened 2023-02-14
fix(get_full_task): rework check for task's episode and sequence

robin-vagner:
**Problem**
The logic of `get_full_task` when getting the episode and sequence did not take entities of type `Scene` into account.

**Solution**
Changed the logic to make it work.
I also changed the condition for assets entities. I did not use `assets_service.is_asset` because it seemed like a costly function, but if you would find it more explicit to use it I can modify my commit.


EvanBldy:
Thanks @robin-vagner

PR #601, opened 2023-02-14
refactor(services.files): Refactor query for last output files

robin-vagner:
**Problem**
Not really a problem, but `get_last_output_files_for_entity` and `get_last_output_files_for_instance` could be simplified a little bit.

**Solution**
We refactored the functions to be a bit more concise.


EvanBldy:
Thanks @robin-vagner

PR #602, opened 2023-02-14
Add possibility to get timespents for a specific date range.

robin-vagner:
**Problem**
We needed a way to get a timespent for a specific date range, that could overlap on two months.

**Solution**
We added a general `time-spents` route, that accepts a `start_date` and an `end_date` argument.
The current `time-spents/<date>` route is unchanged except for its resource name, to which we prefixed `Date`.

These modifications were made for both `/data/persons` and `/data/user` routes.


EvanBldy:
Thanks @robin-vagner

PR #603, opened 2023-02-14
Add route to guess an entity from a given path.

robin-vagner:
**Problem**
We needed a way to create output files from a file or folder, but we also wanted to avoid having the user specify which entity it was, to reduce potential errors.

**Solution**
We added this `guess_from_path` route, that allows us to guess an entity from a path.


EvanBldy:
Thanks @robin-vagner

PR #628, opened 2023-03-23
fix(file_tree_service): Use ilike in guess data from path

CoralieGold:
**Problem**
The tokens' values defined in path can have a different case or format than the values in database.
Example: we can have `model_render` in the path corresponding to a task name `Model Render`.

**Solution**
Using ilike for the name column fixes these differences.


frankrousseau:
Thank you @CoralieGold !

PR #661, opened 2023-06-23
Upgrade libs

EvanBldy:
**Problem**
- Some libs are outdated

**Solution**
- Upgrade libs


frankrousseau:
Some tests are failing. Can you check what is wrong?

PR #667, opened 2023-07-10
[previews] support NTSC frame rate with 3 decimal places

NicoPennec:
**Problem**
Some users use a NTSC camera at 23.976 fps, but currently Zou rounds to 2 decimal places.

**Solution**
Increase the frame rate to 3 decimal places.


NicoPennec:
Some software rounds to 2 digits, others to 3 digits. 
It may not be useful to go up to 3 decimal places...

EvanBldy:
> Thank you for the PR. Could you fix the tests?

I will fix them :)

PR #684, opened 2023-08-17
Update indexer.md

manuelrais:
Fix zou reset-search-index command

**Problem**
The `zou reset-search-index` was written twice.

**Solution**
Removed extra command


frankrousseau:
Thank you @manuelrais !

PR #694, opened 2023-09-06
Add a github action to push on PyPi

nledez:
**Problem**
We want to simplify the process to push packages on PyPi

**Solution**
Use a Github action to build package and push it on PyPi


frankrousseau:
Thank you @nledez

PR #696, opened 2023-09-11
[metadata] improvements

frankrousseau:
**Problem**

* It is possible to create a metadata descriptor with the wrong type.
* There is no type for tag lists.

**Solution**

* Add the `taglist` type.
* Ensure that a correct type is given on updates and creations


EvanBldy:
Thanks @frankrousseau

PR #716, opened 2023-10-02
[WIP] Add prometheus /metrics

nledez:
**Problem**
We need metrics

**Solution**
Add prometheus metrics endpoint


EvanBldy:
- disallow the /metrics for unauthenticated users 
- add an environment variable to enable/disable prometheus metrics

PR #725, opened 2023-10-16
Simple AD fixes

Plyrolith:
**Problem**
Simple AD authentication is using different attributes for user creation (sAMAccountname) and authentication (CN).
CN in combination with `full_name` only resolves properly when user admins follow a schema using spaces, which is not guaranteed and should therefore be avoided.
Addionally: When using LDAP_GROUP, there was a bug in the condition structure that led to None being iterated.

**Solution**
sAMAccountname should be used when logging in, since the attribute is stored as `desktop_login`.
This provides consistency and guarantees successful authentication, no matter which user naming schema is used.


EvanBldy:
Hi @Plyrolith, thanks for your PR :)

PR #727, opened 2023-10-17
Allow 'id' field for file trees

Plyrolith:
**Problem**
In many productions, entity (display) names need to remain dynamic, since assets may be under heavy development and shots may be constantly re-arranged.
If file structures are handled by pipeline applications and there's no need to make the tree human-readable, the generated Postgres UUIDs are a better option for creating persistent file/folder names.

**Solution**
I included the field parameter in all relevant folder generator functions and made sure its default remains "name".
I made sure the dash is not caught by `slugify` when updating the variables.

~~I also removed a redundant `slugify` in `get_file_name_root`. The `slugify` caught spelling variations in (non-variable) static file name parts; since this was not happening for static folder path parts I assumed this is unwanted. I would expect the "style" to only affect variables coming from the DB; any statically defined text within the file tree should not be altered.~~ This needs to be restored for file names without touching the UUID dashes.

Modification: Before file names are slugified and stylized, all UUIDs are stored (re pattern findall). In the end, their modified versions are replaced with the original ones to ensure intact UUID format.

### DIFF
--- a/zou/app/services/file_tree_service.py
+++ b/zou/app/services/file_tree_service.py
@@ -31,7 +31,10 @@ from zou.app.services.exception import (
     TaskNotFoundException,
 )
 
-ALLOWED_FIELDS = ["short_name", "name", "number"]
+ALLOWED_FIELDS = {"short_name", "name", "number", "id"}
+UUID_PATTERN = re.compile(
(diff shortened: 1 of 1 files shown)


EvanBldy:
Hi @Plyrolith, 
Thanks for your PR
It breaks the tests can you fix that ?

Plyrolith:
Yes, I already know what’s the issue. Will fix it tomorrow!

PR #728, opened 2023-10-19
use cn for simple ad user creation and auth

Plyrolith:
**Problem**
The previous change proved to not work with all AD server setups in production.
Binding with *sAMAccountName* lead to problems depending on the directory.

**Solution**
Use CN for binding again. Store CN as `desktop_login` on user creation. Although CN is not unique by design, it can be considered unique in the Zou context since it is appended to a single LDAP_BASE_DN. Ambiguous requests are practically impossible and it should be safe for user identification. In the worst case, email is always available as identifier.
Regular (non-SIMPLE) AD remains untouched.
I thoroughly tested this against our production directory servers this time.


EvanBldy:
Thanks @Plyrolith

PR #731, opened 2023-10-27
display_aspect_ratio of video file may not exist

mathbou:
**Problem**

In some cases, the `display_aspect_ratio` property is not set in video files, and FFprobe does not compute it, leading to the crash of the `get_movie_display_aspect_ratio` method.

**Solution**

Compute the aspect ratio using the video resolution.

### DIFF
--- a/zou/utils/movie.py
+++ b/zou/utils/movie.py
@@ -168,10 +168,20 @@ def get_movie_fps(movie_path=None, video_track=None):
 
 def get_movie_display_aspect_ratio(movie_path=None, video_track=None):
     """
-    Returns movie display aspect ratio (width / height).
+    Returns movie display aspect ratio, if display aspect ratio
+    is not set, fallback on width / height.
     """
-    width, height = get_movie_size(movie_path=movie_path, video_track=video_track)
-    ratio = width / height
+    if video_track is None:
+        video_track = get_video_track(
+            movie_path, "get_movie_display_aspect_ratio"
+        )
+    ratio = 1
+    if video_track is not None:
+        try:
+            width, height = video_track["display_aspect_ratio"].split(":")
+        except KeyError:
+            width, height = video_track["width"], video_track["height"]
+        ratio = float(width) / float(height)
     return ratio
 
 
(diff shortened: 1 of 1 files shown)


EvanBldy:
Hi @mathbou, do you have a sample of a video size that have "display_aspect_ratio" not set ? Or can you send the stacktrace that leads to the crash ? 
We use DAR to get the real ratio (not the ratio from the video resolution that can be not real because of black borders) so I think we should use the resolution to get the ratio only if DAR is not provided.

mathbou:
I used [this sample](https://www.pexels.com/video/a-boy-wears-a-virtual-reality-headset-3209828/) video for my test.

EvanBldy:
Effectively, the DAR is not provided in this video. 
But we need to use it if possible (if not we have tiles that have wrong ratio), so I propose you something like that :
```
def get_movie_display_aspect_ratio(movie_path=None, video_track=None):
    """
    Returns movie display aspect ratio (width / height), if display aspect ratio
    is not set use width / height.
    """
    if video_track is None:
        video_track = get_video_track(
            movie_path, "get_movie_display_aspect_ratio"
        )
    ratio = 1
    if video_track is not None:
        if "display_aspect_ratio" in video_track:
            width, height = video_track["display_aspect_ratio"].split(":")
        else:
            width, height = video_track["width"], video_track["height"]
        ratio = float(width) / float(height)
    return ratio
```

mathbou:
Yes, that could be a good middle ground. However, I'm wondering in which case DAR does not match the width and height ? You mentioned black borders, but even with a file having borders, I always had matching DAR and resolution 🤔

EvanBldy:
It depends if the file is with incrusted black borders or not. 


> Yes, that could be a good middle ground. However, I'm wondering in which case DAR does not match the width and height ? You mentioned black borders, but even with a file having borders, I always had matching DAR and resolution 🤔

For example cinemascope :) 
You have some samples examples here from ffmpeg : https://samples.ffmpeg.org/aspect_samples/

mathbou:
Okay, I get it. If the video is anamorphic, pixels aren't squared and there is a mismatch 👍

EvanBldy:
Thanks @mathbou

Issue #732, opened 2023-11-02
Mobile devices do not login, or do not resolve after login.


flinfo:
**Context**

Zou version: 0.17.23 and 0.17.45
Zou installation type: self-hosted

**Describe the bug**
Mobile phones (Android and Apple) do not resolve page after (assumed successful) login. Only white screen displayed.

**Expected behavior**
After login, one would expect to enter into kitsu and a page be displayed, eg. open-productions.

**Screenshots**
n/a - purely white screen

**Additional context**
We have 2 kitsu installs, which work fine and operate as expected when using Mac or Windows devices. When trying to access either of them using a mobile device, the initial login screen appears as expected. After entering email and passwords correctly, the screen goes white and that is as far as it ever goes.
The url that appears after (apparent) successfull login is: https://egwebsite.com/login?redirect=/
If I refresh the page, or type a direct url eg: https://website.com/open-productions, then the login page appears again.

Upgrading to the latest Zou and Kitsu do not change results (ie. same behaviour)
Adding or removing 2FA does not change the situation.
Tried multiple Android and IOS devices.

gunicorn_access.log:
`127.0.0.1 - - [02/Nov/2023:02:48:45 +0000] "GET //config HTTP/1.0" 200 45 "https://website.com/login?redirect=%2F" "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Mobile Safari/537.36"
127.0.0.1 - - [02/Nov/2023:02:48:45 +0000] "GET //config HTTP/1.0" 200 45 "https://website.com/login?redirect=%2F" "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Mobile Safari/537.36"
127.0.0.1 - - [02/Nov/2023:02:49:07 +0000] "POST //auth/login HTTP/1.0" 200 1679 "https://website.com/login?redirect=%2F" "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Mobile Safari/537.36"
[...]

PR #735, opened 2023-11-03
[login] on login set cookies for mobile browsers (chrome)

NicoPennec:
**Problem**
- We don't set cookies on login for chrome mobile browsers. (related to #732)

**Solution**
- Set cookies on login for chrome mobile.


EvanBldy:
Thanks !

PR #754, opened 2023-12-19
Fix update filter

NicoPennec:
**Problem**
- On updating filters, we can't unlink a filter from a filter group.

**Solution**
- Allow updating a filter without a group ID to unlink it from the group.

Example of payload:
```js
{
  name: "test",
  search_query: "test",
  search_filter_group_id: null
}
```


EvanBldy:
Hi @NicoPennec,
Thanks for the PR, the problem is that "name" and "search_query" can't be None when updating a search filter. 
I have done a fix to only allow search_filter_group_id to be None.

PR #762, opened 2024-01-26
Push stats information on database upgrade

frankrousseau:
**Problem**

There is no way to know how many Kitsu installations exist.

**Solution**

When a database upgrade is performed, it pushes anonymized statistics to the CGWire telemetry service to allow to know: 

- The number of studios using Kitsu
- The number of users
- The number of previews posted
- The number of comments posted


EvanBldy:
Thanks !

PR #765, opened 2024-02-01
Add Sentry config for Kitsu

NicoPennec:
**Problem**
- We need Sentry configuration data for Kitsu provided by the API.

**Solution**
- Add new environment variables related to Kitsu Sentry and send them via the `/config` endpoint if set.


EvanBldy:
thanks !

PR #766, opened 2024-02-01
[breakdown] Add a CLI command to reset data.

frankrousseau:
**Problem**

Sometimes the breakdown statistics data stored at the task level are unsync.

**Solution**

Add a CLI command to clean the breakdown stats set at the task level.


EvanBldy:
thanks

PR #769, opened 2024-02-08
File routes improvements

frankrousseau:
**Problem**

* Single elements querying for file management are not cached
* There is no route to grab all output files for a given project

**Solution**

* Cache `get_software`, `get_output_type`, `get_working_file`, `get_output_file`
* Add a `/data/projects/<project_id>/output-files` route


EvanBldy:
Thanks!

PR #770, opened 2024-02-09
[previews] Allow to set previews automatically as main preview

frankrousseau:
**Problem**

When the productions progress, thumbnails are not updated automatically.

**Solution**

If the option is set at the production level, set the last published preview as the thumbnail of the related entity and store the last preview file id at the task level.


EvanBldy:
Thanks!

PR #771, opened 2024-02-09
[bot] introduce new bots

EvanBldy:
**Problem**
- There's no possibility to add bots for connecting with the API.

**Solution**
- Add that possibility : a bot is just a person with a flag "is_bot" at True, an expiration date (can be set to None, for a perpetual API token) and an access_token.


zou/app/blueprints/crud/person.py
```diff
@@ -133,15 +148,20 @@ def post_update(self, instance_dict):
         instance_dict["departments"] = [
             str(department.id) for department in self.instance.departments
         ]
+        if "expiration_date" in data:
+            instance_dict[
+                "access_token"
+            ] = persons_service.create_access_token_from_instance(
```
frankrousseau, line 154: This function naming is weird. Can you explain what it does?

zou/app/blueprints/user/resources.py
```diff
@@ -25,7 +24,6 @@ class AssetTasksResource(Resource):
     Return tasks related to given asset for current user.
     """
 
-    @jwt_required()
```
frankrousseau, line 28: Why do you remove authentication from all these routes?

zou/app/models/person.py
```diff
@@ -56,14 +41,28 @@
...
+
+
 class Person(db.Model, BaseMixin, SerializerMixin):
     """
     Describe a member of the studio (and an API user).
     """
 
     first_name = db.Column(db.String(80), nullable=False)
     last_name = db.Column(db.String(80), nullable=False)
-    email = db.Column(EmailType, unique=True)
```
frankrousseau, line 66: Why do you remove unicity of emails?

zou/app/blueprints/crud/person.py
EvanBldy: It creates the access token for the bot from its instance

zou/app/blueprints/user/resources.py
EvanBldy: It's because it's directly set in all routes of the blueprint : https://github.com/cgwire/zou/pull/771/files#diff-773e47b9349c32eebf93133c5858bfb2383e2f8f29c56d61e2dbecb079a63307R110
This blueprint is blocked for bots

PR #776, opened 2024-02-19
 Add project_id to output files events

frankrousseau:
**Problem**

Output file events don't give the project ID which leads to unnecessary requests.

**Solution**

Add the project id information from the entity information.


EvanBldy:
Thanks !

PR #823, opened 2024-06-21
Feature/asset type short name

scottmcdonnell:
**Problem**
Add a short_name to the entity type so that Asset Types can have short names. This is useful for naming conventions of workfiles - we can shorten "Environments" to "ENV" to match our workfile names. Brings it in line with other models that have short_name fields.


EvanBldy:
Hi @scottmcdonnell, 
Thanks for your PRs, we will merge them this week (we need to do a release before).

PR #825, opened 2024-06-26
bugfix order of migrations

scottmcdonnell:
**Problem**
2 PRs with migrations have the same down_revision

**Solution**
update the down_revision so the migrations will upgrade correctly.


EvanBldy:
Hi @scottmcdonnell, I already have fixed that in this PR : https://github.com/cgwire/zou/pull/826
So I will close this PR, thanks anyway :)

PR #827, opened 2024-06-28
add a command to renormalize preview files

EvanBldy:
**Problem**
- When a movie preview file fails to be normalized in nomad for example it's impossible to relaunch the normalization.

**Solution**
- Add a new command to renormalize movie preview files.

### DIFF
--- a/zou/app/blueprints/previews/resources.py
+++ b/zou/app/blueprints/previews/resources.py
@@ -338,8 +338,6 @@ class CreatePreviewFilePictureResource(Resource, ArgsMixin):
             preview_files_service.prepare_and_store_movie(
                 preview_file_id, uploaded_movie_path, normalize=normalize
             )
-            preview_file = files_service.get_preview_file(preview_file_id)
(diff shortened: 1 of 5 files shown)


zou/app/blueprints/previews/resources.py
```diff
@@ -338,8 +338,6 @@ def save_movie_preview(
             preview_files_service.prepare_and_store_movie(
                 preview_file_id, uploaded_movie_path, normalize=normalize
             )
-            preview_file = files_service.get_preview_file(preview_file_id)
-            tasks_service.update_preview_file_info(preview_file)
```
frankrousseau, line 342: why do you remove the metedata from the preview file?

EvanBldy: Because it's already done in the function before : https://github.com/cgwire/zou/blob/master/zou/app/services/preview_files_service.py#L305

EvanBldy: So it was done twice

PR #834, opened 2024-07-26
[quotas] Allow to count quota based on done date

frankrousseau:
**Problem**

It's not possible to compute quotas on the done date instead of the feedback date.

**Solution**

* Add a `done_date` on tasks
* Change the boolean `weighted` query flag to a string variable deciding of the count mode.
* Adapat quoat functions to be able to switch between count modes


EvanBldy:
Thanks!

PR #841, opened 2024-08-20
Various improvements

frankrousseau:
**Problem**

* There is no way to check if the DB is initialized via the command line.
* Some TV Paint extensions are missing
* It's not possible to update shots via the `shots/shot_id` route
* Some routes are non accessible to non-admin users while there is no functional reason for that: `/entities` and `projects/poject_id/tasks`
* `projects/poject_id/tasks` cannot be filtered by episode

**Solution**

* Add a `is-db-ready` command to the CLI to check if the DB was initialized.
* Add TV Paint extensions
* Allow to update shots via PUT requests on the `shots/shot_id` route
* Open access to `/entities` and `projects/poject_id/tasks` by filtering on project teams (assigned tasks for vendors)
* Allow to filter tasks from `projects/poject_id/tasks`  via an `episode_id` parameter 
* Allow to filter tasks from `tasks`  via an `episode_id` parameter

### DIFF
--- a/zou/app/blueprints/crud/entity.py
+++ b/zou/app/blueprints/crud/entity.py
@@ -11,6 +11,7 @@ from zou.app.models.entity import (
     EntityLink,
     EntityConceptLink,
 )
+from zou.app.models.project import Project
 from zou.app.models.subscription import Subscription
 from zou.app.services import (
     assets_service,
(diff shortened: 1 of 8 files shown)


EvanBldy:
Thanks !

PR #854, opened 2024-09-05
Various fixes

frankrousseau:
**Problem**

* Project deletion fails when there is a descriptor tied to a department.
* It's possible to create asset types with the same name if cases are different.
* Bump.sh doesn't accept the open API documentation due to curl examples restrictions.

**Solution**

* Remove descriptor <-> department links before deleting the project.
* Refuse the asset type creation with name equals whether letters are uppercased or not.
* Remove curl examples and send to the Zou documentation.


EvanBldy:
Thanks

PR #862, opened 2024-09-23
Add setting to use hours instead of days for durations and estimations

MaximeGrosmaire:
**Problem**
We have a mix of artist that can work more or less daily on a project so it's important for us to be able to estimate in hours.

**Solution**
I've added an organization-wide setting to toggle duration and estimates in hours. 

This change is required for these kitsu changes to work: https://github.com/cgwire/kitsu/pull/1535

Should I also be committing the migration file as part of this pr?


frankrousseau:
Thank you for the PR. It requires some data migration. We will handle it a little bit later to ensure nothing is breaking.

PR #864, opened 2024-09-27
Feature/test performance - pytest speed up

scottmcdonnell:
**Problem**
pytest all tests running very slow as all database tables are created and dropped for each test

**Solution**
move the table creation and drop to once per class setUpClass and tearDownClass.
Set up a Sqlalchemy session in each task setUp that is rolled back in tearDown.

Result: big speed increase in running tests.

### DIFF
--- a/tests/auth/test_auth_route.py
+++ b/tests/auth/test_auth_route.py
@@ -24,17 +24,10 @@ class AuthTestCase(ApiDBTestCase):
             "email": self.person_dict["email"],
             "password": "secretpassword",
         }
-        try:
-            self.get("auth/logout")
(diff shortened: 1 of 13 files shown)


frankrousseau:
Hello @scottmcdonnell, 

Thank you for this very helpful contribution. Could you make all tests green? If not we'll have a look at it. It seems most of them are ok.

scottmcdonnell:
I took this speed up from another project of ours and adjusted it from pytest fixtures to your unit test classes. Unfortunately, I don't have the bandwidth at the moment to debug the few tests that are failing. If you don't get to it I can look at it later this month.

PR #866, opened 2024-10-08
Various improvements

frankrousseau:
**Problem**

* More allowed extensions are needed: .kra, .sai, .sai2 (Krita and Paintool).
* Difficulty per task can't be set.
* Upcoming preferences are required: dark theme by default, publish by default for artists, and compression bitrates.
* It's not possible to mark all user notifications as read.
* More notifications are required: read/unread, watching/not watching.

**Solution**

* Add extensions to the allow list.
* Add a difficulty field (integer, 1 to 5) to the Task model.
* Add related preference field on the Organisation and Project models. 
* Add migration file.
* Add a route to mark all notifications as read.
* Join subscription info to the notification query and add support for filtering on the read field and the fact there is a related subscription or not.


EvanBldy:
Thanks

PR #876, opened 2024-10-18
[person] allow a person in PROTECTED_ACCOUNTS to update her informations

EvanBldy:
**Problem**
- A person in PROTECTED ACCOUNTS can't change information in her profile and password.

**Solution**
- Fix that.


PeteDraper:
I love the solution :D

PR #886, opened 2024-11-20
Added missing field on present_minimal method of Person

stevenpince:
**Problem**
The user service runs persons_service.get_persons as minimal if the given user doesn't have Manager permissions.
The problem is that the present_minimal method in /zou/app/models/person.py#179 doesn't return the "type" field.
This broke our pipeline, as the output of the /data/Persons webhook was different between Artists and Managers

**Solution**
Added the "type" field to the representation for non-manager users.


EvanBldy:
Thanks @stevenpince

PR #913, opened 2025-01-24
various improvements

EvanBldy:
**Problem**
- There's no "force" parameter to delete files on the filesystem for preview file routes.
- For /data/projects/<project_id>/asset-types/<asset_type_id>/assets/new "is_shared" arg is at True by default. 
- When importing a task type CSV it's not possible to import difficulty field.
- For status automation when importing the last revision there's an exception when there are no preview files. 
- Some requirements are outdated. 

**Solution**
- Add a "force" parameter to delete files on the filesystem for preview file routes.
- For /data/projects/<project_id>/asset-types/<asset_type_id>/assets/new set "is_shared" arg at False by default. 
- When importing a task type CSV allow to import the "difficulty" field.
- For status automation when importing the last revision check before if there are no preview files. 
- Upgrade outdated requirements.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     flask-migrate==4.1.0
     flask-socketio==5.5.1
     flask==3.1.0
-    gazu==0.10.23
+    gazu==0.10.24
     gevent-websocket==0.10.1
     gevent==24.11.1
     gunicorn==23.0.0
@@ -110,7 +110,7 @@ monitoring =
 lint =
     autoflake==2.3.1
     black==24.10.0
-    pre-commit==4.0.1
+    pre-commit==4.1.0
 
 [options.entry_points]
 console_scripts =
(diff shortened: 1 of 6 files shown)


zou/app/services/tasks_service.py
```diff
@@ -329,9 +329,9 @@ def get_shot_tasks_for_episode(episode_id, relations=False):
     Get all shots tasks for given episode.
     """
     query = _get_entity_task_query()
-    Sequence = aliased(Entity, name="sequence")
-    query = query.join(Sequence, Entity.parent_id == Sequence.id).filter(
-        Sequence.parent_id == episode_id
+    sequence = aliased(Entity, name="sequence")
+    query = query.join(sequence, Entity.parent_id == sequence.id).filter(
+        sequence.parent_id == episode_id
```
frankrousseau, line 334: Here we are aliasing a model. That's why it should keep an uppercase at the beginning.

EvanBldy: Ah yeah good point, will remove this commit !

PR #921, opened 2025-02-25
[tasks] Add more filters to the open task route

frankrousseau:
**Problem**

It's not possible to filter open tasks on the studio, the department, and multiple assignees.

**Solution**

Allow `studio_id`, `department_id` and several `person_id` separate by commas.


zou/app/models/task.py
```diff
@@ -47,6 +47,8 @@ class Task(db.Model, BaseMixin, SerializerMixin):
     last_comment_date = db.Column(db.DateTime)
     nb_assets_ready = db.Column(db.Integer, default=0)
     data = db.Column(JSONB)
+    nb_drawings = db.Column(db.Integer, default=0)
```
EvanBldy, line 50: The migration for this is missing.

zou/app/models/project.py
```diff
@@ -165,6 +165,7 @@ class Project(db.Model, BaseMixin, SerializerMixin):
     is_publish_default_for_artists = db.Column(db.Boolean(), default=False)
     hd_bitrate_compression = db.Column(db.Integer, default=28)
     ld_bitrate_compression = db.Column(db.Integer, default=6)
+    is_drawing_mode = db.Column(db.Boolean(), default=False)
```
EvanBldy, line 168: The migration for this is missing.

EvanBldy (approved):

PR #941, opened 2025-04-11
patch error when reply is None in get_last_notifications()

mdk-man:
**Problem**
In some situation user can't view their notifications filtered by task type = ALL
api/data/user/notifications is returning a HTTP 500, as a workaround user have to filter by Task Type, one after another, to understand where the notification comes from.

**Solution**
around line 1459 a check is made for 'reply is not None', meaning that None is an accepted value. 
But few lines earlier reply.get("mentions", []) is called which will ultimately throw an AttributeError when reply is None ('NoneType' object has no attribute 'get'). 
Solution is to wrap everything into a if/else statement to define default values.

### DIFF
--- a/zou/app/services/user_service.py
+++ b/zou/app/services/user_service.py
@@ -1452,12 +1452,17 @@ def get_last_notifications(
                 ),
                 None,
             )
-            reply_mentions = reply.get("mentions", []) or []
-            reply_department_mentions = (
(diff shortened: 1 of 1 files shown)


EvanBldy (approved):

EvanBldy:
Hi @mdk-man,
Thanks for your PR :)

PR #948, opened 2025-04-23
Budget management

frankrousseau:
**Problem**

There is no way to make budget predictions in Kitsu.

**Solution**

- Add a salary scale and salary fields at the person level (seniority, position, and daily salary).
- Add a budget table and a budget entry table to set up versioned and projected budgets


EvanBldy:
Thanks @frankrousseau

PR #959, opened 2025-05-22
[projects] Add a route to get all time spents statistics for a given project

frankrousseau:
**Problem**

There is no way to get an aggregation of the project time spents. 

**Solution**

Add a route to get aggregated time spents by departments and persons for a given project.


EvanBldy:
Thanks !

PR #975, opened 2025-07-12
Update index.md to remove duplicate SECRET_KEY definition

CaskAle:
I don't think it is needed to define SECRET_KEY twice in /etc/zou/zou.env

**Problem**
I don't think it is needed to define SECRET_KEY twice in /etc/zou/zou.env.  It appears twice in the example.

**Solution**
Remove one line of SECRET_KEY declaration in /etc/zou/zou.env section of doc.


frankrousseau:
Thank you @CaskAle!

PR #981, opened 2025-08-26
[openapi] Fix asset resources descriptors

frankrousseau:
**Problem**

* OpenAPI docstrings are not properly set for asset resources.

**Solution**

* Adapt to OpenAPI v3 and move POST information to the request body instead of form data.


EvanBldy:
Thanks @frankrousseau

PR #988, opened 2025-09-29
Various improvements

frankrousseau:
**Problem**

* A customer cannot reply to a customer from its studio
* There is no event when an asset `ready_for` field is changed via an automation.

+ Open API docs improvements

**Solution**

* Allow clients from the same studio to reply to each other (work for clients without studios too) in the reply resource
* Emit an `asset:update` event when the asset is modified through an automation.


EvanBldy:
Thanks @frankrousseau

Issue #966, opened 2025-06-11
Minor development environment setup doc improvements


hww-nikhilshankarnoble:
Hi CG-Wire team! 

I setup a local development environment for Zou for the first time yesterday to test the events system, following the documentation at https://zou.cg-wire.com/development/

While it was very easy to setup, there were a few very small things that caught me out initially that I wanted to suggest as improvements that might make it easier for future users. I'm happy to make these changes myself in a PR but I don't believe the documentation is open-source anymore? I might just be blind, but I couldn't find it under the organisation.

* When installing Meilisearch, it instructs you to pull `v1.8.3` but then runs with `v1.5` - which causes it to pull again (and the v1.8.3 image is never used). I had some small bumps setting up the indexer (listed below) and this gave me a little confusion over the version I should be using.
* The indexer was originally not working for me when following the setup instructions, I was getting issues that the indexer wasn't initialised. I eventually stumbled across https://zou.cg-wire.com/indexer/#refreshing-indexes which led me to run the `DB_PASSWORD=... INDEXER_KEY=... zou reset-search index` command that fixed this for me. Not sure if I did something wrong during the setup here but might be worth a mention!
* In order for the `tests_search.py` module to succeed, the `INDEXER_KEY=meilimasterkey` environment variable should be set for the `py.test` command. It would be good to note that this is required - and that `ffprobe` must be installed for the movie tests to pass.

Thanks!
Nikhil

EvanBldy:
Thanks for the report @hww-nikhilshankarnoble. 

I will try to improve the documentation soon!

frankrousseau:
The documentation was improved.

PR #993, opened 2025-10-03
Move Meilisearch install instructions from index.md to indexer.md

harywilke:
**Problem**
The installation instructions Meilisearch are spread between index.md and indexer.md. Also the instructions on index.md were incomplete because they fail to mention the environment variables needed to make the indexing work, and the instructions on index.md fail to have you create the meilisearch user.

**Solution**
Merge the two separate instructions into one and move it to the indexer.md page.
Include a link to the indexer.md instructions in index.md where the previous instructions were.

As this is an optional install, it's not necessary to be included with the main install instructions.
This is a better solution as having the instructions in one place means less to maintain.
This addresses the second bullet point for [issue #996](https://github.com/cgwire/zou/issues/966)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -231,51 +231,12 @@ article](https://www.techandme.se/performance-tips-for-redis-cache-server/).
 
 ### Set up the indexer (optional)
 
-Create a Meilisearch user:
+To allow full-text search, Kitsu relies on an Indexing engine. It uses the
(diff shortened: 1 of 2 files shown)


frankrousseau:
Thank you for the contribution. We have to extract the right commit to merge it. It will require more time than usual.

PR #1000, opened 2025-11-27
[playlists] add column created_by

EvanBldy:
**Problem**
- We need this created_by column. 

**Solution**
- Add created_by column + permissions


NicoPennec (changes requested):
The DB migration is missing.

Issue #1003, opened 2025-12-15
Feature: enable all entity task types for schedule


mxatmx:
Hello, after being irritated that I couldn't see all entity task types on the production schedule page, I engineered a fix. Just opening this issue so I can close it with a PR.

Basically the same as #837 

I'm also opening an issue and PR in the Kitsu frontend repository that enables filtering what entity task types you would like to see at runtime. That is here: [https://github.com/cgwire/kitsu/pull/1921](https://github.com/cgwire/kitsu/pull/1921)

PR #1005, opened 2025-12-17
feat: enable all entities for schedule display #1004

mxatmx:
**This PR is recreated from a previous because I had to change the head branch.**

Problem
Closes https://github.com/cgwire/zou/issues/1003

Solution
Now all entity task types are sent to the frontend for display and possible filtering.

I should emphasize: BOTH THIS PR AND THE PR ON THE FRONTEND KITSU REPO (https://github.com/cgwire/kitsu/pull/1924) MUST BE APPROVED FOR THIS FUNCTIONALITY TO WORK PROPERLY. Without the frontend changes, the necessary data is not displayed or available to filter properly.


mxatmx:
> Hello can remove the commit related to the deletion? The deletion should occur only after the rights validation.

Yes, I can move the cascade deletion to after rights validation, my apologies for the incorrect placement. I was having an error where a foreign key constraint was blocking deletion of task types, because of lingering scheduleItem records. I'll use the pre_delete hook for this instead.

PR #1041, opened 2026-03-22
feat: Add entity type verification to status automation processing and…

mxatmx:
…corresponding tests for sequence, episode, and mismatch scenarios.

I PRed a sister commit to the Kitsu repo.

### DIFF
--- a/tests/services/test_status_automation_service.py
+++ b/tests/services/test_status_automation_service.py
@@ -1,8 +1,12 @@
 # -*- coding: UTF-8 -*-
 from tests.base import ApiDBTestCase
 
+from zou.app.models.status_automation import StatusAutomation
+from zou.app.models.task import Task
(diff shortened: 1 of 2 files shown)


frankrousseau:
Applied the normalization fix in 5dcf2e66:

```python
wanted = (automation.get("entity_type") or "").lower()
if wanted == "asset":
    if not assets_service.is_asset_type(entity_type):
        return
elif entity_type["name"].lower() != wanted:
    return
```

Regarding the `Entity` import: it is actually used at line 80 (`self.asset = Entity.get(self.asset.id)`) to refresh the asset instance after the automation runs. So the import is needed and stays.

All 6 automation tests pass (including the new sequence, episode, and mismatch tests).

frankrousseau:
Merged via the command line.

PR #1047, opened 2026-04-07
perf: reduce memory consumption across services

nledez:
**Problem**
Several SQLAlchemy patterns cause excessive memory usage: Task.assignees is eagerly loaded via selectin on every query (70-80% don't need it), get_comments() triggers N+1 queries fetching persons one by one, backup_service loads all preview files into memory at once, and playlists_service fetches full ORM objects when only 2 fields are needed.

**Solution**
  - Fix N+1 in get_comments() by batch-fetching persons with get_persons_by_ids()
  - Stream preview files in backup_service with yield_per(500)
  - Use with_entities(id, extension) in playlists_service instead of full PreviewFile objects
  - Replace manual if key not in dict patterns with defaultdict in projects_service and time_spents_service
  - Flush index documents in batches in index_service instead of accumulating indefinitely
  - Change Task.assignees from lazy="selectin" to default lazy loading, add explicit selectinload only in the 3 bulk access sites (CSV export, schedule service, deletion service)

### DIFF
--- a/zou/app/blueprints/export/csv/base.py
+++ b/zou/app/blueprints/export/csv/base.py
@@ -40,14 +40,14 @@ class BaseCsvExport(Resource):
         self.prepare_import()
         try:
             self.check_permissions()
-            csv_content = []
-            csv_content.append(self.build_headers())
-            results = self.build_query().all()
-            for result in results:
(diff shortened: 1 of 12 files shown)


frankrousseau (commented):
See comments

PR #1096, opened 2026-06-01
[tasks] Block task type deletion with clean 400 when still linked

frankrousseau:
**Problem**
- Deleting a task type still attached to projects or scheduled items hit a foreign-key IntegrityError, returning an opaque 500.
- The previous `pre_delete` only purged ScheduleItems and ignored ProjectTaskTypeLink rows entirely.

**Solution**
- Without `force`, refuse the delete with a 400 listing the blockers (schedule items and/or projects).
- With `force=true`, purge ScheduleItem and ProjectTaskTypeLink rows before removing the task type.


frankrousseau:
Reopening as a main->main PR bundling other pending changes.

PR #1099, opened 2026-06-02
[previews] Fix upload error messages, frame extraction and eraser rendering

frankrousseau:
## Problems

- Preview upload failures were misleading: a revision conflict returned a bare 400, and movie processing failures were always labelled "Normalization failed" even when normalization was disabled
- Single-PNG-frame extraction produced inconsistent output because ffmpeg was not told it was overwriting an existing frame
- The eraser annotation field was not handled by the renderer, so erased strokes were not applied

## Solutions

- Make the revision-conflict message actionable, and report genuine server-side movie processing failures as 500 with the cause in `data.reason` (400 kept only for bad uploads)
- Pass `update=1` to ffmpeg when writing a single PNG frame
- Handle the eraser annotation field in the renderer

### DIFF
--- a/tests/tasks/test_preview_processing_failure.py
+++ b/tests/tasks/test_preview_processing_failure.py
@@ -1,0 +1,69 @@
+import os
+
+from unittest import mock
+
+from tests.base import ApiDBTestCase
+
+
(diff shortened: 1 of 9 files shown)


nledez (approved):
LGTM

PR #1103, opened 2026-06-06
Feat/preview missing status

nledez:
**Problem**
  - renormalize-movie-preview-files failures left PreviewFile.status stuck on processing, so Kitsu kept showing affected previews as processing forever even though the
  renormalize pass had given up.
  - The single broken status conflated two very different cases: a movie that exists but cannot be normalized (encode/ffmpeg error, transient storage hiccup) versus a movie
  whose source binary is simply gone from object storage. Operators could not target the irrecoverable rows without retrying the recoverable ones too.
  - sync-full-files re-downloaded every preview every time, including rows known to be unusable, wasting bandwidth and time on previews that will never play.

**Solution**
  - renormalize-movie-preview-files now marks failed rows on exit: source-binary-missing → new missing status, other failures → broken. The 3-preview reproducer that prompted
  this no longer stays as processing.
  - New internal status missing added to PreviewFile.STATUSES. No DB migration needed — ChoiceType is enforced at ORM level only.
  - EXPOSE_MISSING config flag (env var, default false) gates the API exposure of the new status. With the default, missing is folded back to broken in serialize(), present()
  and present_minimal() so existing API/Kitsu clients see no change. This flag is intentionally temporary: it gives us time to (1) announce the new status publicly, (2) ship the
   matching Kitsu UI/UX changes, and (3) let third-party clients adopt. Once Kitsu speaks missing natively, EXPOSE_MISSING will be removed and missing will be surfaced
  unconditionally.
  - renormalize-movie-preview-files gains a --all-missing flag (sibling of --all-broken / --all-processing); --all-broken stays broken-only, --all-missing is the new operator
  escape hatch to retry rows whose source has since been restored.
  - sync-full-files gains --skip-broken and --skip-missing flags (default off, so behavior is unchanged) to let operators exclude unusable rows from large pulls.
  - New service helper preview_files_service.set_preview_file_as_missing(); internal queries that filter ("broken", "processing") now include "missing" so it is treated the same
   as broken for "needs attention" purposes.
  - Tests added for the missing→broken serialization remap (both EXPOSE_MISSING values) and for the --all-missing filter behavior.

### DIFF
--- a/specs/configuration.md
+++ b/specs/configuration.md
@@ -59,6 +59,12 @@ All configuration is in `zou/app/config.py`, read from environment variables.
 | `SENTRY_DSN` | | Sentry error tracking |
 | `PLUGIN_FOLDER` | | Path to plugins directory |
 
+## Preview files
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `PREVIEW_SAVE_SOURCE_FILE` | false | Keep the uploaded source movie alongside the normalized preview |
+
 ## LDAP / SAML
 
 | Variable | Default | Description |
--- a/tests/misc/test_commands.py
+++ b/tests/misc/test_commands.py
@@ -163,3 +163,42 @@ class RenormalizeMoviePreviewFilesTestCase(ApiDBTestCase):
 
         self.assertIn("Local copy of source movie is missing or", output)
         mock_prepare.assert_not_called()
+
+    def test_source_missing_marks_preview_as_missing(self):
+        from zou.app.models.preview_file import PreviewFile
+
+        with patch.object(
+            file_store, "exists_movie", return_value=False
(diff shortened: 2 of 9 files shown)


zou/app/config.py
```diff
@@ -80,6 +80,7 @@
     os.getenv("THUMBNAIL_FOLDER", os.path.join(os.getcwd(), "previews")),
 )
 PREVIEW_SAVE_SOURCE_FILE = envtobool("PREVIEW_SAVE_SOURCE_FILE", False)
+EXPOSE_MISSING = envtobool("EXPOSE_MISSING", False)
```
frankrousseau: Why do you add this flag? It's ok to exposte missing to everyone.

PR #1118, opened 2026-07-01
Skip bot email validation on update

ivancea:
# Problem

Bot user updates fail with `Email already in use` when the bot shares an email with a regular user. Making impossible to update any bot field after creation.

## Reproduction

1. Create a bot using the same email as an existing user (Default behavior from Kitsu)
2. Edit only the bot name
3. Save

Before: update fails with `Email already in use`.
After: bot update succeeds, matching bot creation behavior.

## Solution

Ignore email check when updating a bot


frankrousseau:
Thank you!

PR #1176, opened 2026-07-23
Speed up schedule version copy on large productions

NicoPennec:
**Problem**
- Copying task links into a schedule version 500'd past ~10900 tasks (Postgres 65535 parameter limit) and timed out over 60s on large productions.

**Solution**
- Copy links and assignees in SQL (INSERT ... SELECT) and return a lightweight count instead of the discarded payload.
- Portable to PostgreSQL 12, self-copy rejected, apply-to-production lightened the same way.

### DIFF
--- a/tests/services/test_schedule_service.py
+++ b/tests/services/test_schedule_service.py
@@ -1,13 +1,20 @@
+from datetime import datetime
+
 from tests.base import ApiDBTestCase
 
+from zou.app import db
 from zou.app.models.entity import Entity
 from zou.app.models.milestone import Milestone
 from zou.app.models.production_schedule_version import (
     ProductionScheduleVersion,
+    ProductionScheduleVersionTaskLink,
+    ProductionScheduleVersionTaskLinkPersonLink,
 )
+from zou.app.models.task import Task
 from zou.app.services import schedule_service
 from zou.app.services.exception import (
     ProductionScheduleVersionNotFoundException,
+    WrongParameterException,
 )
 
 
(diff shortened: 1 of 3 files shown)


frankrousseau (approved):
Reviewed and validated, with a focus on the API contract and the consumers.

- The INSERT ... SELECT approach is the right one. I verified the ON CONFLICT targets against the actual constraints (`production_schedule_version_task_link_uc`, person link composite PK), and the asymmetric assignee refresh (all links from production, source-matched links only from a version) is correct.
- The response shape change is safe: kitsu awaits and discards all three responses, gazu has no wrappers for these routes, and the developer docs do not document them.
- Test coverage is thorough, including the self-copy assignee wipe scenario and PG 12 portability.

One non-blocking suggestion: the self-copy guard compares raw strings. The target id comes from the URL path while the source id is the serialized lowercase one, so an uppercase UUID in the path bypasses the guard while the SQL still matches (Postgres compares UUIDs case-insensitively), leading to the exact assignee wipe the guard prevents. Passing `production_schedule_version["id"]` from the route (already fetched and normalized), or normalizing both sides in the service, closes it.

Minor nits, also non-blocking:
- `copied_count` comes from a separate count query, which can diverge from the copy under concurrent writes; `rowcount` on the insert would be exact and one query cheaper.
- `apply-to-production` still emits one `task:update` per task. Pre-existing, but the next bottleneck at this scale.

NicoPennec:
Self-copy guard: good catch. Fixed by passing the already-fetched, normalized `production_schedule_version["id"]` from the route, with a route-level regression test. (see new commit [b23754e](https://github.com/cgwire/zou/pull/1176/commits/b23754e41102881f76bf7ec4d4f418404e28854d))

On `copied_count`: I made it exact, but not via `rowcount`. `rowcount` is -1 for `INSERT ... SELECT ... ON CONFLICT` (an earlier iteration shipped that, and CI caught it), so I count the insert's `RETURNING` rows instead. This is exact under concurrency and uses one fewer query.

On `apply-to-production` events: agreed it's the next bottleneck, but the per-task `task:update` events go to the frontend over Redis, and there is no batched task-event shape it handles today, so collapsing them requires a coordinated frontend change (or a maintainer call on dropping the per-task `ApiEvent` persistence). Keeping it as a separate follow-up rather than changing an endpoint this PR does not otherwise touch.

PR #1184, opened 2026-08-03
[comments] Let a mentioned person answer in the task they were pulled into

tinnhcgi:
### Problem

Mentioning someone in a task comment sends them a notification and an email (`notifications_service.create_notifications_for_task_and_comment`), so they open the task — and then cannot answer. Posting goes through `check_task_action_access`, which only allows assignees, supervisors of the task's department, managers and clients. The mention brings them into the conversation and the permission check keeps them silent.

For an artist this is a daily annoyance: the only ways out today are to add them as an assignee (which puts the task in their My Tasks and their quota) or to raise their role. Neither is what "please take a look at this" means.

### What this changes

`check_task_mention_access` grants the narrowest right that makes a mention useful: the person was named in a comment of *that* task, either directly (`comment_mentions`) or through one of their departments (`comment_department_mentions`).

- Team membership is checked first, so the vendor test reads the project role rather than the global one. Vendors stay out.
- `CommentTaskResource.post` only falls back to it when `task_status_id` equals the task's current status — a mentioned person can write, but cannot move the task.
- Previews still go through their own routes, which keep the full `check_task_action_access`.
- Replies carry neither a status nor a preview, so the check alone is enough in `ReplyCommentResource.post`.

Nothing changes for anyone who already had access.

### Companion UI change

The comment box is hidden client-side by `isCommentingAllowed`, so this alone is not visible to users. The matching frontend change is cgwire/kitsu#2169. **Please do not merge one without the other** — the UI patch alone would show a comment box the API rejects.

### Status

Running in production since 2026-08-03 on 1.0.62 with the frontend counterpart: mentioned artists can now answer in the task, the status selector is hidden for them, and preview upload is still refused.

I have not added tests yet — `tests/comments/test_route_comments.py` looks like the right home for them. Happy to write them, and equally happy to rework the approach if you would rather express this differently (a project flag, or folding it into `check_task_action_access` instead of a separate check).


tinnhcgi:
Updated: the check now also covers mentions made **inside a reply**.

The first version only read `comment_mentions` and `comment_department_mentions`, which turns out to miss most real cases — a reply keeps its mentions inside the `replies` JSONB, and replies are where people actually name each other once a conversation is going. On our production database the split is 1148 comments carrying a reply that mentions a person, and 96 carrying a department mention in a reply.

The JSONB side is matched with containment (`replies @> '[{"mentions": ["<id>"]}]'`), verified against real rows before shipping. The frontend counterpart in cgwire/kitsu#2169 was updated the same way.

Running in production with this version since 2026-08-04.

frankrousseau:
Don't you think they should have access to the reply instead of posting a new comment?

tinnhcgi:
Honestly no, and the reason is the shape of the conversation rather than the
permission itself.

The case this comes from is a supervisor pulling extra people into a task: a
pipeline TD, the comp, the onset guy, whoever holds the missing piece. That conversation
runs for the life of the task, not for one exchange.

With reply-only, what happens in practice is: someone is mentioned, answers
once, goes back to their own work. They investigate and come back a day and two
revisions later with what they found. By then the comment they were mentioned in
is stale context, sitting behind newer comments and a couple of status changes,
and it is the only place they are allowed to write. So they ping a supervisor out
of band asking to be re-mentioned somewhere current, just to be able to speak.
The mention becomes a one shot ticket instead of participation, and the
workaround costs more manager time, not less.

The grant is also already manager controlled. Nobody gets it by accident:
someone who already has full access to the task has to deliberately name them.
Making them re-earn that on every revision does not add safety, it adds
friction. And what a mention still cannot do is the part that matters for
integrity: no status change, no preview upload. It grants text on the task, and
that is the whole scope.

For context, I run the studio this came from, and the patch has been in
production with us. Reply-only is closer to the behaviour my
artists asked me to fix in the first place.

If the concern is that this is too permissive as an upstream default, I would
rather put it behind a setting (instance or project level, default off) than
narrow it to replies. Say the word and I will add the flag, plus tests in
tests/comments/test_route_comments.py.

One more thing, since it frames this patch. I think the real shape of the
problem is entity level: everyone assigned to any task on a shot or asset is
already that shot's crew, and today they cannot talk to each other across task
types. check_task_action_access already reaches into the entity for the creator
case, so this would be a sibling of that branch rather than a new model, with
vendors excluded explicitly via has_vendor_permissions and the rule limited to
shots and assets. Mentions would then go back to being what they should be,
pulling in someone from outside the shot. I would open that separately rather
than growing this PR. Happy to write it up if you are interested.

frankrousseau:
Ok we will take it. Can you give us more information about your studio? Name, type of projects?

tinnhcgi:
Sure, and thanks for taking it.

I am the CEO and VFX supervisor at Step One (stepone.art), a VFX studio in Vietnam working on theatrical features for Vietnamese and Korean productions. We run Kitsu deeply integrated into our internal pipeline, which is where these patches came from.

While I have you: cgwire/kitsu#2169 is the UI counterpart to this one and is still open. Now that the API side is in, that PR is what actually makes the change visible to users. Happy to rebase or adjust it if anything needs changing.

